PR #4, opened 2017-10-12
Dev

russgove:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [ ]
| New sample?      | [ ]

Implement Grouping in ListView

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "prepublishOnly": "gulp"
   },
   "dependencies": {
+    "lodash": "^4.17.4"
   },
   "devDependencies": {
     "@microsoft/sp-build-web": "~1.3.0",
--- a/src/controls/listView/IListView.ts
+++ b/src/controls/listView/IListView.ts
@@ -18,6 +18,14 @@ export interface IListViewProps {
    */
   viewFields?: IViewField[];
   /**
+   * The fields you want to group your list view by
+   */
+  groupByFields?: string[];
+   /**
+   * The fields you want to sort your list view by
+   */
+  sortByFields?: string[];
+  /**
    * Boolean value to indicate if the component should render in compact mode.
    * Set to false by default
    */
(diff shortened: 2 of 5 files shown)


estruyf:
Thanks @russgove will try to test it out this week.

estruyf:
@russgove I did some tests and there are various problems so I cannot merge it.

- debugger statements
- findIndex method cannot be found because you switched to lodash
- grouping not correct
- sorting not correct

estruyf:
I rewrote grouping for the ListView control. If you want to test it out, it is available in the DEV branch: https://github.com/SharePoint/sp-dev-fx-controls-react/tree/dev

russgove:
Ill check it out  today

russgove:
Still Some Issues. I have my test data set up like this:
[image: Inline image 1]

Want To group by Field1 and Field 2 so i set it to
const groupByFields: string[] = ["ListItemAllFields.Field1",
"ListItemAllFields.Field2"];

When I load the test webpart I see this
[image: Inline image 2]

Doc B1 is in the B2 group.
Also, I think it should sort ascending by whatever we group on and be
collapsed (or let user specify)

estruyf:
Something went wrong with your images, can you re-upload them, please.

Sorting is achieved per group.

estruyf:
I think I found the bug. Did an update, if you have time to check it out, that would be wonderful.

russgove:
Ok, I tested it out and the items are showing in  the correct groups now.
But  the groups start out in reverse  Order:


[image: Inline image 1]



and when I click on the title to sort , The Groups get sorted:
[image: Inline image 2]

The way I think it should work is that the groups are 'constant'.  sorting
columns does not change the sequence of the groups, just the order of the
items within the groups.

estruyf:
The order of the groups should now be fixed. Also added the ability to tell how you want to sort the groups: ascending or descending.

Issue #8, opened 2017-11-02
Pre-selection of items


joelfmrodrigues:
#### Category

[X ] Question

#### Expected / Desired Behavior / Question
Does the control support pre-selection of items

#### Observed Behavior
Hi, I'm trying to pre-select items on the ListView component, but I can't find a way to do that. My main problem is that if I update the state, the render method fires again and any selection on the list is lost.
I'm not sure if the component was designed to allow this (checked the source code and I believe it was not) but is there any way to overcome this?

#### Steps to Reproduce
Add the ListView component to a page
Add a button to the page that calls a function where the status is updated
Select some items on the list
Click the button to update the status
The list renders and the previous selection is lost

Thanks!

joelfmrodrigues:
Hi again,

I have found a simple way to implement this in the component.
I will submit a PR soon

Thanks,
Joel

joelfmrodrigues:
Closing this issue as PR was merged

PR #9, opened 2017-11-03
Select default items

joelfmrodrigues:
| --------------- | ---

| New feature?    | [X ]

| Related issues?  | mentioned in #8 

Added new property to allow an array of default items to be passed to the component. 
Added a new function that will select the items passed by default.
Calling the function on componentDidUpdate, so the items are selected and persisted if the render method is executed

This allow for simple scenarios like just selecting items by default when the page renders, but also more complex scenarios, like cascading lists (select items on the first list that will trigger selection of items on the second list)

### DIFF
--- a/src/controls/listView/IListView.ts
+++ b/src/controls/listView/IListView.ts
@@ -14,6 +14,10 @@ export interface IListViewProps {
    */
   items: any[];
   /**
+   * The items to select by default.
+   */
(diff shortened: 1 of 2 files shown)


joelfmrodrigues:
Closing PR as I can't resolve merge issues....Github failed to fork dev branch when I forked the repo so I need to wipe it out and fork again

PR #10, opened 2017-11-03
Select default items

joelfmrodrigues:
| Q               | A
| --------------- | ---

| New feature?    | [ X]

| Related issues?  |  mentioned in #8 

Added new property to allow an array of default items to be passed to the component.
Added a new function that will select the items passed by default.
Calling the function on componentDidUpdate, so the items are selected and persisted if the render method is executed

This allow for simple scenarios like just selecting items by default when the page renders, but also more complex scenarios, like cascading lists (select items on the first list that will trigger selection of items on the second list)

### DIFF
--- a/src/controls/listView/IListView.ts
+++ b/src/controls/listView/IListView.ts
@@ -19,10 +19,6 @@ export interface IListViewProps {
    */
   items: any[];
   /**
-   * The items to select by default.
-   */
-  selectIems?: any[];
-  /**
    * The fields you want to view in your list view
    */
   viewFields?: IViewField[];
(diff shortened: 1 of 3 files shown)


estruyf:
Thanks for this PR. Just one remark. Wouldn't it be better to pass an array of item IDs instead of the actual item? That would make it easier to set the default selected items by user input.

joelfmrodrigues:
@estruyf I will try to change this over the next days and update the PR

joelfmrodrigues:
@estruyf  I have been unable to pick this up due to work...
Are you making the assumption that the items will always have an Id property? Or am I missing something?

joelfmrodrigues:
Hi @estruyf , I have now updated this to select items by index as it looked like the most sensible option to me so that it mirrors Details List functionality. Let me know if you still prefer an Id based implementation.
Additionally, I have added a tslint.json file at the root to allow the TSLint extension for VSCode to work with the rules on the config folder

estruyf:
Thanks, this is now merged to the dev branch.

PR #48, opened 2018-02-19
Fix for sorting listview items

estruyf:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]

Fix for the `ListView` control when sorting with selected items.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Releases
 
+## 1.2.2
+
+**Fixes**
+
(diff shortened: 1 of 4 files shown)


BubaGuM:
Hi,

This fix has introduced an issue where sorting no longer works if the optional selection property is not set for ListView. Throws an undefined error when being referred to in line 353 const selection = this._selection.getSelection();

estruyf:
Thanks for reporting this @BubaGuM. I'll create an issue for this and will be fixed in the next release. You can check the status here: #65

PR #50, opened 2018-03-07
Lookup Field Renderer Optimization

AJIXuMuK:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | no

Lookup field rendering using `FieldRendererHelper` was optimized.
Previously, `FieldRendererHelper` was sending request to get url of Display form for the list that is referenced in the lookup first and only then rendered the Lookup.
Because of that there was a lag before lookup values are rendered in the cell.
But the Display form url is needed only if a user clicks on the Lookup value. It's not needed to render the values itself.
So, the code was changed to request the url from `FieldLookupRenderer` itself at the moment when a user clicks on the value.


estruyf:
Thanks for the update. Will be included in the next release.

PR #61, opened 2018-04-16
Taxonomy Picker Control

spdavid:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [x ]
| New sample?      | [ ]

Fixed the Taxonomy Picker with AutoComplete and ability to pick terms. Also AnchorId can be set if you want to pick further down in the term tree.


spdavid:
It cant find my TaxonomyPicker.Module.scss file according to circleci build test. Any Ideas?

estruyf:
@spdavid many thanks, this just got merged. I did some changes/fixes to the code.

- Fixed the suggestions so that it only returns terms which are not yet selected
- There was an issue when providing a term set name instead of an ID. I changed the code so that it is now supported to use the name of the term set. 
- Removed some `console.log`

This control is now available in the `v1.3.0` beta release. Are you also working on the documentation for the control?

spdavid:
Thought i got all those console.logs. Nice with the fixes. yes docs will be coming next.

PR #92, opened 2018-07-12
Implemented demo users for PeoplePicker when on LocalWorkbench

Octavie:
Implemented demo users for PeoplePicker when on LocalWorkbench
Implemented interface IPeoplePickerUserItem when provisioning PeoplePicker with user data

| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [ ]
| Enhancement      | [X ]

Added some demo users when using the control locally with the local workbench. You can now also test the event methods, such as selectedItems.

Also implemented the IPeoplePickerUserItem interface that is used when provisioning the control instead of using a undefined-any-array-element.


AsishP:
@Octavie , Great work. Sorry I have been away for few days, so couldn't provide more inputs.

Octavie:
@asishp Thanks. No worries, had fun doing so. Now I am on holidays, so after that I pick up some more cool stuff to do. Cheers!

PR #105, opened 2018-08-06
Reference ouifr statically instead of dynamic.

wobba:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]

Refernce Office UI Fabric statically instead of dynamically to ensure only what is used is loaded/included.


estruyf:
Thanks @wobba for the fix!

PR #126, opened 2018-09-12
Fix to lookupDispFormUrl

prokach:
| Q               | A
| --------------- | ---
| Bug fix?        | [#X ]
| New feature?    | [ ]
| New sample?      | [ ]

lookupDispFormUrl was fixed in order to provide a correct behavior when user clicks on a lookup field

### DIFF
--- a/src/controls/fields/fieldLookupRenderer/FieldLookupRenderer.tsx
+++ b/src/controls/fields/fieldLookupRenderer/FieldLookupRenderer.tsx
@@ -128,7 +128,7 @@ export class FieldLookupRenderer extends React.Component<IFieldLookupRendererPro
         //
         if (this.state.dispFormUrl) {
             this.setState({
-                lookupDispFormUrl: `${this.state.dispFormUrl}&ID=${lookup.lookupId}&RootFolder=*&IsDlg=1`,
+                lookupDispFormUrl: `${this.state.dispFormUrl}?ID=${lookup.lookupId}&RootFolder=*&IsDlg=1`,
                 hideDialog: false
             });
         }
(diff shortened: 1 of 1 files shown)


AJIXuMuK:
I'm not sure about that fix.

Most probably it crashes the functionality.
@prokach could you please describe what exactly wasn't working before that fix?

prokach:
@AJIXuMuK Redirection to a Display Form of an item, which is selected in a lookup field was crashed.

prokach:
@AJIXuMuK Unfortunately, it didn't work even if `dispUrlForm ` is not provided by user in `props`.
What other situations do you mean? Usually `dispUrlForm` is provided in `props` as `.../DispForm.aspx`, isn't it?

AJIXuMuK:
Hi @prokach ,

There are multiple ways to use `FieldLookupRenderer` and, as a result, work with `dispUrlForm` value.
1. Use `FieldLookupRenderer` control directly and provide `dispUrlForm` in React `props`. In that case `dispUrlForm` can basically contain anything that a user wants.
2. Use `FieldLookupRenderer` control directly but instead of providing `dispFormUrl` pass `fieldId`. In that case the `dispFormUrl` will be generated using `SPHelper.getLookupFieldListDispFormUrl` method. **NOTE:** this options as well as an option 3 works in Field Customizer but won't work in Web Part.
3. Use `FieldRendererHelper` class that automatically "understands" what type of control to use in the specific List Field.

Options 2 and 3 uses the next format for `dispFormUrl`: `${w.Url}/_layouts/15/listform.aspx?PageType=4&ListId=${f.LookupList}`. In that case we need to add additional parameters using ampersands, not question mark as we already have query parameters in the url.

If you look at OOTB implementation of the lookup popup it uses the same format for DisplayForm: listform.aspx?PageType=4.

That's why I've bit updated your fix, that is definitely needed as well, to take into consideration all 3 scenarios of usage and both formats to provide `dispFormUrl`.

Let me know if it makes sense.

prokach:
@AJIXuMuK Thanks a lot for your answer. Now it's really clear.

PR #130, opened 2018-09-14
Fix the people picker freezing issue

### LINKED ISSUE #128: PeoplePickerComponent doesn't work properly if try to add one more user
#### Category
[ ] Enhancement

[ X ] Bug

[ ] Question

When I select some user in PeoplePicker and trying to find another one in order to add one more user, a suggestion list is empty instead of showing a list of users which are related to what I've started to type.

#### Observed Behavior
The suggestion list always has to contain a number of users which have the same email/name(or ever role) that I started to type - started to find

#### Steps to Reproduce
1. Click on user field area
2. Type name or email of user1
3. Select the user from a suggestion list
4. Click on user field area again
5. Type name or email of user2
The suggestions list is empty

#### Submission Guidelines
It looks like, that the issue occurs because __listContainsPersona_ function which is using for filtering of suggestion list works with primaryText propery, but the items of users don't have this property(see for example on creation of a user item in __thisLoadUsers_ function)

primaryText properties here are **undefined**
```
  private _listContainsPersona = (persona: IPersonaProps, personas: IPersonaProps[]): boolean => {
    if (!personas || !personas.length || personas.length === 0) {
      return false;
    }
    return personas.filter(item => item.primaryText === persona.primaryText).length > 0;
  }
```

It's a peace of __thisLoadUsers_ function. Here is no  **primaryText** defined
  ```
           userValuesArray.push({
                id: item.Id.toString(),
                imageUrl: this.generateUserPhotoLink(item.Email),
                imageInitials: "",
                text: item.Title, // name
                secondaryText: item.Email, // email
                tertiaryText: "", // status
                optionalText: "" // anything
            });
```

Thanks!

gautamdsheth:
| Q               | A
| --------------- | ---
| Bug fix?        | Y
| New feature?    | N
| New sample?      | N
| Related issues?  | fixes #128 , #134 

As mentioned in #128, people picker component freezes when you try to add another user.  This happens in the local as well as SharePoint workbench because `primaryText` is undefined. So, modified it to use `secondaryText`. This fixes the freezing issue and we are able to add multiple users

### DIFF
--- a/src/controls/peoplepicker/PeoplePickerComponent.tsx
+++ b/src/controls/peoplepicker/PeoplePickerComponent.tsx
@@ -322,7 +322,7 @@ export class PeoplePicker extends React.Component<IPeoplePickerProps, IPeoplePic
     if (!personas || !personas.length || personas.length === 0) {
       return false;
     }
-    return personas.filter(item => item.primaryText === persona.primaryText).length > 0;
+    return personas.filter(item => item.secondaryText === persona.secondaryText).length > 0;
   }
 
   /**


estruyf:
Thanks @gautamdsheth appreciate the fix, but implemented another fix myself. The issue why `primaryText` is `undefined` comes from the fact that in the latest versions of Office UI Fabric the properties changed. Before the `NormalPeoplePicker` used `primaryText` where it now uses `text`. All references of `primaryText` had to be updated in the code instead of falling back on `secondaryText`. Although this was a good approach, it does not cover all. Will close this PR without merging it, but you will receive credits in the changelog for this.

Thanks.

gautamdsheth:
Thanks @estruyf  for such a detailed response. I should have investigated it more and should been more careful before sending this PR, apologies.

PR #167, opened 2018-10-30
Update spfx 1.6.0

### LINKED ISSUE #144: spfx v 1.6
#### Category
[ ] Enhancement

[ ] Bug

[x] Question

I am thinking about adding some functionality to the people picker to grab data from the graph. This requires SPFx 1.6 or greater. Also will have to add some webApiPermissionRequests or at least some documentation to set it up if people want to use that functionality.  Perhaps a setting UseGraph or PeopleSource property to define where the data is coming from.

Any reason why we cant upgrade the solution to SPFx1.6?

joelfmrodrigues:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [X]
| New sample?      | [ ]
| Related issues?  | fixes #144 

Update to SPFx 1.6.0
I have tested the controls and all seems to be working fine, but had to change some imports for the PeoplePicker and, despite all seem to be working, I'm not 100% sure of the impact.


joelfmrodrigues:
@estruyf I have removed the explicit dependency on office ui fabric from package.json as per [documentation](https://docs.microsoft.com/en-us/sharepoint/dev/spfx/office-ui-fabric-integration?view=sp-typescript-latest)

estruyf:
@joelfmrodrigues the PR is becoming a bit complicated. Might be better to squash your commits and do a new PR.

About the explicit Office UI Fabric dependency reference. Why would you remove it? This was intentionally added due to the many changes added to OUIF in the property names. Before we were always relying on the OUIF version of the current SPFx project, but now for some controls, we retrieved errors and warnings about the properties. That is why this was added to the package.json file. Best is that the project follows the latest version of SPFx.

joelfmrodrigues:
@estruyf thanks for the feedback, I will add "office-ui-fabric-react": "~5.120.0" back again as that makes sense. It's also the same version as SPFx 1.6.0, so all fine.
Regarding the PR, yes it was a bit messy, sorry for that...The number of silly mistakes just rockets when trying to do things at night. But if you look at the files changed, you will notice that all the changes are to json files and most of them are minimal.

PR #184, opened 2018-11-16
[People Picker] change `titleText` to non-required property

### LINKED ISSUE #175: PeoplePicker titleText property should not be required
#### Category
[ x ] Enhancement

[ ] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: 1.10.0

The `titleText` property of the PeoplePicker (and other controls) should not be required, as is the case with many other third party react input controls. It should be optional. The user should be able to render the control without a title, or perhaps in a table cell where the title is in the header of the column.

#### Observed Behavior
The `titleText` property of the PeoplePicker is required, and if an empty string is given, it renders a default title.

tsekityam:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #175 

Change `titleText` to non-required property, and hide the label if `titleText` is not provide.


estruyf:
Thank you @tsekityam for this change, your change is now merged and will be available in version `1.11.0`.

Issue #14, opened 2017-12-06
New control: Map control


estruyf:
A new control that allows you to include a map into your SPFx solution.

estruyf:
The map control will be available in version `1.11.0`.

![map-control](https://user-images.githubusercontent.com/2900833/49653683-ff67bf00-fa35-11e8-8894-7702315f0aff.gif)

PR #185, opened 2018-11-19
New control - Maps and also adding "ExtensionContext" to PeoplePicker and List Cntrol

anoopt:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [X ]
| New sample?      | [X ]
| Related issues?  | NA

1. Added a new control called "Maps" as a response to #14 

> The current version of the submitted PR, accepts co-ordinates (latitude and longitude) as properties and displays [OpenStreetMap](https://www.openstreetmap.org) for the co-ordinates. 
> The plan is to add "city search" in the upcoming weeks and use Google's geocode service to get co-ordinates for the entered city. This will need API key (which is free only for some queries and after that would need billing enabled for premium services).

2. Added "ExtensionContext" reference in ListPicker and PeoplePicker controls.

> There are times when we would need to use these controls in SPFx extensions as well. Hence I have added the reference to "ExtensionContext" in these controls and the context in these controls can now be of the type "WebPartContext" or "ExtensionContext".

`context: WebPartContext | ExtensionContext;`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6431,12 +6431,14 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "bundled": true,
           "dev": true,
+          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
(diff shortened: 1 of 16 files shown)


anoopt:
Adding more changes - hence closing this PR.

AJIXuMuK:
It’s not necessary to close the PR if you want to make more changes. Just add commits to the branch and push them. All the commits will be automatically added to the PR. 
From other hand, to my mind, it’s better to have separate PR for different functionality. 
In current situation I would create one PR fir `ExtensionContext` and another for Map control.

anoopt:
Thank you Alex for the suggestion.

Issue #97, opened 2018-07-14
People Picker is searching users only from current site and not entire active directory


harshdamaniahd:
I have noticed that people picker is searching users present in current site only . It should fetch all users in active directory
       ```
       <PeoplePicker
                context={this.props.pageContext}
                titleText="People Picker"
                personSelectionLimit={1}
                groupName={''}// Leave this blank in case you want to filter from all users
                showtooltip={true}
                isRequired={false}
                tooltipMessage={"Search for people"}
                selectedItems={this._getPeoplePickerItems}
              />
```

AsishP:
@harshdamaniahd , I suppose we could another property to allow search all Active Directory (or SharePoint User Profiles or Graph to simplify it). 

Currently the scope of the people picker is set to use the site users and not Global user profiles. The idea of doing that is that the control fetches validated users of the site. Is there a particular scenario for using AD or User Profiles?  Would love to know the reason of the usage before coming up with a technical plan.

harshdamaniahd:
@AsishP 
For now there are two use cases,
1. There is an intranet landing page, which has an option to search user. So it's great to use this control as when you type you can see the user.Once selected you can click on search icon , which redirects to delve page giving information about the user selected.
2.There is custom form which basically has columns  like approvers ( here we require this control).

At least for now

AsishP:
@harshdamaniahd , I suppose scenario 2 is a perfect fit for this control where approvers will be users of the site mostly I think. Please let me know otherwise. 

However, in my opinion, the control was not build to handle scenario 1. We are working on a similar scope in #23 (New Search Control) but haven't made much progress on it recently. I could add People search to the existing scope of the search control. Will that work?

harshdamaniahd:
@AsishP  Thanks.
Actually it took me some time to think and I have some questions (My understanding)
In case we are using people picker and permission is managed via Office 365 groups, then in that case the user will never be shown up in results.So i guess the scope of PeoplePicker becomes very limited.Generally a people picker searches all users .

Talking about Search Control , so search will take place only for users and no other content just like people picker right ? 
Yeh..this seems okay as well

But then its better to segregate content search and people search.
This is what I feel.But definitely you guys may have a better understanding.

My only requirement is that user can enter name and it will return result just like people picker :-)

harshdamaniahd:
@AsishP  so any thoughts ?

AsishP:
[...]

PR #196, opened 2018-12-09
Implementing People Picker to search all user profiles and not just current site (#97)

AsishP:
| Q               | A
| --------------- | ---
| Bug fix?        | [X ]
| New feature?    | [X]
| New sample?      | [ ]
| Related issues?  | Implements People search for all tenant users as mentioned in #97 

The items implemented in this PR:
1. Search across All Users instead of only this site by default.
2. Added resolveDelay to allow search of users across the tenant
3. Binding defaultUsers to non existent site users
4. Updated Documentation with details for resolveDelay and added Principal type

### DIFF
--- a/docs/documentation/docs/controls/PeoplePicker.md
+++ b/docs/documentation/docs/controls/PeoplePicker.md
@@ -37,7 +37,8 @@ import { PeoplePicker, PrincipalType } from "@pnp/spfx-controls-react/lib/People
     disabled={true}
     selectedItems={this._getPeoplePickerItems}
     showHiddenInUI={false}
-    principleTypes={[PrincipalType.User]} />
+    principleTypes={[PrincipalType.User]}
+    resolveDelay={1000} />
 ```
 
 - With the `selectedItems` property you can get the selected People in the Peoplepicker :
@@ -73,6 +74,7 @@ The People picker control can be configured with the following properties:
 | showHiddenInUI | boolean | no | Show users which are hidden from the UI. By default these users/groups hidden for the UI will not be shown. | false |
 | principleTypes | PrincipleType[] | no | Define which type of data you want to retrieve: User, SharePoint groups, Security groups. Multiple are possible. | |
 | suggestionsLimit | number | no | Maximum number of suggestions to show in the full suggestion list. | 5 |
+| resolveDelay | number | no | Add delay to resolve and search users | 1000 |
 
 Enum `PrincipalType`
 
@@ -84,5 +86,6 @@ The `PrincipalType` enum can be used to specify the types of information you wan
 | DistributionList |
 | SecurityGroup |
 | SharePointGroup |
+| All |
 
 ![](https://telemetry.sharepointpnp.com/sp-dev-fx-controls-react/wiki/controls/PeoplePicker)
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,17372 @@
+{
+  "name": "@pnp/spfx-controls-react",
+  "version": "1.11.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@microsoft/api-extractor": {
+      "version": "3.2.3",
+      "resolved": "https://registry.npmjs.org/@microsoft/api-extractor/-/api-extractor-3.2.3.tgz",
+      "integrity": "sha1-U/HQb+h8+uRParkrKUH+aaNkhog=",
+      "dev": true,
+      "requires": {
+        "@microsoft/node-core-library": "~0.2.7",
+        "@types/es6-collections": "0.5.29",
+        "@types/fs-extra": "0.0.37",
+        "@types/node": "6.0.62",
+        "@types/z-schema": "3.16.31",
+        "fs-extra": "~0.26.7",
+        "jju": "~1.3.0",
+        "typescript": "~2.4.1",
+        "z-schema": "~3.18.3"
+      }
+    },
+    "@microsoft/decorators": {
+      "version": "1.3.4",
+      "resolved": "https://registry.npmjs.org/@microsoft/decorators/-/decorators-1.3.4.tgz",
+      "integrity": "sha1-zJK+IaaF7MFNOnEKSSRtVGz8J2o=",
+      "dev": true
+    },
+    "@microsoft/gulp-core-build": {
+      "version": "3.0.5",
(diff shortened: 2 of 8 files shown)


sympmarc:
@estruyf - I'd love to test this PR. Any idea when it might end up in a beta?

estruyf:
@sympmarc will try to process this later tonight. There are some changes required before this can be merged.

@AsishP thanks again for the updates. The value of `all` should be `15`, not `13`, but it might be better to not include this. As it could create conflicts if you specify for instance `[User, All]`.

estruyf:
It feels that the control is a mix of everything right now. Will require a bit more time in order to rewrite it. 

Example:
- On initial load, the users are retrieved from the site users endpoint
- On search, the users are retrieved from the tenant

IMO we have to go for a similar implementation like in the property pane control. When a site URL is provided it, it searches against the current site. When the site URL isn't provided, it searches over the whole tenant.

@AsishP @sympmarc thoughts?

sympmarc:
Sounds right to me. Ideally, the property pane and this one should work the same (as I suggested in the associated issue) - and should probably share code, if that works. End result should be a people picker that works like any other people picker.

AsishP:
@estruyf , sorry, yeah it should have been 15 don't know why I had left it to 13, I was testing something and forgot to change it back. Updated it back to 15. 

Regarding the logic, I suppose the code is doing as stated in your example. It loads the users from default site on load and then searches from tenant on Inputchange of people picker control. Below the logic in the code at present, please let me know if anything is amiss. Would love to fix any gaps to make it efficient and better. Please suggest.

1. The users are prefilled from site on load for existing site users and userid for existing users and put into the state.
2. Regarding default users, the users are filled from site users and if not present, they are pulled from tenant and updated to state for the control.
3. For non-existent users, search from tenant users (got the idea from property control implementation approach) on InputChange method and then update the state with new users if not present. 
In step 3, since there is a delay for search to return users so had to add a resolveDelay for 1 sec to allow the filter to work. Might there is a better approach here.

Thoughts, suggestions?

estruyf:
@AsishP the option should be at the developer from which source he/she wants to retrieve the data. I like the prefilling idea, but I think it's unnecessary. If the user wants to retrieve data from the tenant, the prefilling should be skipped, as data should come from the tenant search.

When the developer wants to limit data only from the current site, the tenant search should be skipped.

So for your suggestions:
1. Don't feel that we need prefilling, or only when the dev wants to receive data from the current site.
2. Should be defined by the dev from which scope he wants to retrieve data.
3. ...

I think it might be better to completely redevelop the control. Just like @sympmarc said to have the same behavior as in the property pane control.

estruyf:
@AsishP and @sympmarc the control has completely been rewritten. When the `webAbsoluteUrl` property is provided, the people search will be done on the site level. When not provided, this will be at the tenant level.

The code is also simplified + removed a lot of unused state properties. Let me know what is missing or needs to be changed.

New beta version is available for testing.

AsishP:
Ok, let me check the property pane control in detail then to see the flow. I did give it a quick check earlier but not a detailed study. To confirm, this is my understanding on the flow right now.
1. Developers could choose between All Tenant, All Site or Site Group.
2. For All Tenant users, there will be no auto suggestions till search is done for first three characters. The match will be done after third character and fill the state with matches so to speed up future filtering.
3. Will keep the same pre-fill users logic, as the user could be at site or tenant but giving preference to site.

@sympmarc, please give some time for the above to be implemented. @estruyf, I see some conflicts in my PR, so just confirming are these changes with regards to this PR or something else.

AsishP:
Sorry, seems my message came later after @estruyf message. Thanks @estruyf, for the update. Please let me know if I should close this PR.

estruyf:
@AsishP yes, you should best get the latest DEV version, as a lot has been replaced, rewritten, ... 

About point 2, this is indeed a limitation of the API, but don’t feel that this is an issue. As the most of the time you will search, and the proposed once will not be used. 

About point 3, pre-fill logic is implemented if you are retrieving from the site/group. 

We can keep the PR open to discuss it, or we could do it in a new issue. Your call.

ManiKum:
@estruyf , with beta build, the search is happening on entire tenant (as we were expecting). However, the selected item's properties is returning loginname in place of user id:

image

Is this change expected?

sympmarc:
Definitely an improvement, @estruyf and @AsishP! Thanks for digging into this.

Getting the claim back for the id does make some sense, since the selected user may not be in the User Information List yet. (In the previous iteration, since the only users returned were from the current site, they were guaranteed to be in the UIL.)

I think we need to then ensure the user(s) are added to the UIL with a call something like this:

``` typescript
  private _getPeoplePickerRequestedBy(items: any[]): void {
    var updateRequest: IShipmentRequest = this.state.newRequest;
    sp.web.ensureUser(items[0].secondaryText).then((user) => {
      updateRequest.ShipmentRequestor.id = user.data.Id.toString() as string;
      updateRequest.ShipmentRequestor.title = user.data.Title;
      this.setState({
        newRequest: updateRequest
      });
    });
  }
```

estruyf:
@ManiKum when you do the tenant search, you indeed get the user claim. As the API doesn't return the user ID of the UIL as @sympmarc explained. We could indeed add an implementation as @sympmarc provided, or we could also leave it up to the developer. IMO it all depends on what you want to do eventually with the selected user object.

About the `defaultSelectedUsers` property`, can you explain this a bit more? How it currently works is that it validates against the API if the user exists, if he/she is retrieved, they will be added to the control as a selected user.

sympmarc:
It might be a nice option to add `ensureUser` as a parameter? If that was `true`, then then the picker *could* return the local user object. If `false`, it would return the people picker object.

Not sure how best to implement this, but it seems to make sense to tuck it in somehow.

estruyf:
@sympmarc I actually like that idea. Let me try to implement this.

estruyf:
@sympmarc and @ManiKum the `ensureUser` ability has been included to the control. By default, this setting is `false` and returns the user login name. If set to `true`, it will validate the user against the current site and return its ID instead. A new beta is out and available to test.

sympmarc:
@estruyf - In my initial testing, this is working great. Thanks for adding `ensureUser`.

estruyf:
Thanks @sympmarc for testing it out. Will release the public version later this week.

AsishP:
Hi @estruyf , I tried to merge the changes into this PR but there were too many conflicts and didn't know if I was merging it correctly. So I think it is safe for me to close this PR and we go with your commits. Thanks, please let me know.

estruyf:
Yes, we can close this PR.

Issue #190, opened 2018-11-29
iconFieldName issue


Sodininkas:
#### Category
[ ] Enhancement

[ ] Bug

[X ] Question

#### Version

Please specify what version of the library you are using: [     1.10.0    ]

#### Question
How to properly use iconFieldName, in ListView. There are no documentation, just how to add default icon - "serverRelativeUrl". As i understood it should pick file path, as in listview.js - "item[iconFieldName]" - guess file extensions and show that icon. But in reality non of this works. I want to show different icons based on item file type. I can do this in other ways, but why this functionality exist than, no documentation

PR #245, opened 2019-02-28
Update ListView.md: iconFieldName description

PzKfWg:
I had a hard time understanding how to use iconFieldName and its effect. It was not clear to me by the original description that it required the property name of the items with the path of the file and that it would automagically create the entire column, header and field rendering. I had to read the source code to figure it out.

Added precision to the iconFieldName property implementation

Related to issue  #190


estruyf:
Thank you very much @PzKfWg for the documentation update. I have merged this into the DEV branch.

Issue #17, opened 2017-12-11
FileType icon and .pdf


Leif-Frederiksen:
Thank you for reporting an issue, suggesting an enhancement, or asking a question. We appreciate your feedback - to help the team understand your
needs please complete the below template to ensure we have the details to help. Thanks!

**Please check out the [documentation](https://github.com/SharePoint/sp-dev-fx-controls-react/wiki) to see if your question is already addressed there. This will help us ensure our documentation is up to date.**

#### Category
[ ] Enhancement

[X ] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [   1.0.0-beta.8      ]

_If you are not using the latest release, please update and see if the issue is resolved before submitting an issue._

#### Expected / Desired Behavior / Question
When giving af path that ends in .pdf I would expect to get a PDF icon.

#### Observed Behavior
No application specific icon is given - just the default icon (white shet of paper like thing).

#### Steps to Reproduce
Create a webpart and render <FileTypeIcon path="something.pdf">

estruyf:
@Leif-Frederiksen when the control was released, Office UI Fabric didn't contain a PDF icon yet. Apparently, it has it right now, so will check if we can use it in the control set.

estruyf:
@Leif-Frederiksen my mistake, it seems that there is still not a PDF image icon. Only the font icon, which should be working fine.

So if you are using it as follows it should work:

```
<FileTypeIcon type={IconType.font} path="https://contoso.sharepoint.com/documents/filename.pdf" />
```

At the moment it does not work with the `IconType.image` option.

PR #246, opened 2019-02-28
Adding PDF icon support to iconFieldName

PzKfWg:
I realize that it may not be the cleanest approach and would not be shocked if it is not accepted, but here is a workaround that I propose to deal with PDF icons which don't exist in the "image" flavour of FileTypeIcon. I find it annoying that PDF files have the empty icons.
Ideally, I could have access to FileTypeIcon's private _getFileExtension method for a more robust assessment of the PDF extension.

| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [X ]
| New sample?      | [ ]
| Related issues?  | related to #17

PDF files will show an emulated PDF icon instead of the generic file icon.


estruyf:
Office UI Fabric now has a PDF icon. Will add a CDN fallback for this.

Screenshot 2019-03-28 20 17 53

PR #268, opened 2019-04-03
Calculates richtext toolbar top. Fixes #265.

### LINKED ISSUE #265: RichText toolbar appears on top of page
#### Category
[ ] Enhancement

[x] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [1.13.0-beta]

The floating toolbar should appear next to the RichText control. Ideally, the user should be able to hide or move the toolbar out of the way, but it should appear as close to the control as possible or at least in the viewport.

Also, if there are multiple RichText controls, the toolbar should appear next to the control that has the focus/cursor.

I think it is reasonable to require that only one RichText control is in edit mode at any given time to make the toolbar 'stick' to the input. Ideally, with multiple controls, the toolbar should move to there the focus is.

#### Observed Behavior
The toolbar appears at the top of the page regardless of where the RichText control is placed. Scrolling up to the top may be required is the RichText control is too far down the page.

#### Steps to Reproduce
Just place enough divs above the RichText control to push it down to the bottom of the page.

hugoabernier:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #265 

Added code to calculate rich text toolbar position.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6711,12 +6711,14 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "bundled": true,
           "dev": true,
+          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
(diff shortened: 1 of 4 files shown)


estruyf:
Thanks @hugoabernier, I have successfully merged it. 

About the name convention of private variables. I would like to have them without `_` as the editors are smart enough. I tend to follow the following rules: https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines. For now I kept it. Will have to write it down so that new components and updates should all be aligned with the same naming conventions.

hugoabernier:
Thanks for pointing this out, I'll update the code when I submit the next update

PR #271, opened 2019-04-08
SecurityTrimmedControl - Extended to specific folder or item for remote list or lib

Varuuna:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [ X ]
| New sample?      | [ ]
| Related issues?  |

This PR adds 2 optional properties to the SecurityTrimmedControl; folderPath and itemId that can be used when checking permission against a remote list or library. If either is present, we'll check the user's permission against a specified folder or item instead of the list itself.
A new function 'determineResource' is used to determine the url we send with our API call. It will prioritize 'itemId' first, if present, then 'folderPath', then finally 'remoteLibOrListUrl'.


estruyf:
Thanks @Varuuna, I have merged your changed with some modifications to it. I have added two new enum properties to specify the type of security you want to check: `remoteListItem` and `remoteFolder`.

PR #284, opened 2019-04-26
Fix is month picker visible not respected

### LINKED ISSUE #283: DateTimePicker does not respect isMonthPickerVisible prop
#### Category
[ ] Enhancement

[x ] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: 1.13.0

Setting the isMonthPickerVisible prop to false should make the month picker not appear.

#### Observed Behavior
Month picker always appears irrespective of the value of the isMonthPickerVisible prop.

#### Steps to Reproduce
Setting isMonthPickerVisible to false as below has no effect
```typescript
<DateTimePicker
  label="MyLabel"
  isMonthPickerVisible={false}
  onChange={(value) => console.log("DateTimePicker value:", value)}
/>
```

robert-lindstrom:
| Q               | A
| --------------- | ---
| Bug fix?        | [X]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #283

Fixes problem with DateTimePicker not respecting the isMonthPickerVisible prop. This PR fixes that by passing on the isMonthPickerVisible prop to the underlying DatePicker

### DIFF
--- a/src/controls/dateTimePicker/DateTimePicker.tsx
+++ b/src/controls/dateTimePicker/DateTimePicker.tsx
@@ -219,9 +219,9 @@ export class DateTimePicker extends React.Component<IDateTimePickerProps, IDateT
           <div className={styles.time}>
             <div className={styles.picker}>
               <HoursComponent disabled={disabled}
-                              timeConvention={timeConvention}
-                              value={hours}
-                              onChange={this.dropdownHoursChanged} />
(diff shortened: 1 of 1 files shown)


estruyf:
@robert-lindstrom thanks for this fix, it will be included in `1.8.2`

PR #285, opened 2019-04-27
LT localization

ltdu:
Updated LT translations for new controls:
* DateTime picker tranlation is based on CLDR: http://cldr.unicode.org
* RTE - on OOB SharePoint modern Text "web part" 

Additionally, I restructured localization file to follow mystring.d.ts file. This should help once additional properties will be needed for new/updated controls.


estruyf:
Thanks @ltdu for these updates

PR #286, opened 2019-04-28
Fix list item attachments

### LINKED ISSUE #282: listItemAttachments render Issue
#### Category
[ ] Enhancement
[x] Bug
[ ] Question

Please specify what version of the library you are using: [ **1.13.0** ]
Screenshot_4

**Expected / Desired Behavior / Question**

In the **listItemAttachments** control,  at the very end when you delete a file it should render again.

**Observed Behavior**
The  **listItemAttachments**  control doesn't render fine after the click of deleting a file. 

**Steps to Reproduce**
Delete a file and follow the steps. At the and after the file has been deleted, the component won't render fine.

**Submission** Guidelines
I think is a problem with the **DocumentCardPreview**

Thanks!

ltdu:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #282

This fixes #282 by changing how attachment files and previews are loaded. Additionally, it improves attachment related methods in SPService. 

There are couple of things to keep in mind:
* because this changes public API it may require major version increase. ```SPService.getListItemAttachments()``` used to return ```null``` if no attachments there found. Now it returns ```[]```. But this is up to debate as I don't know whether anyone actually uses them directly.
* I'm pretty sure check for existing attachment didn't work before. Now it does and this control replaces existing file with new version. This may be unexpected to some users. Or not...
* I don't use this control in any of my projects so if messed up something let me know :)

### DIFF
--- a/src/controls/listItemAttachments/ListItemAttachments.tsx
+++ b/src/controls/listItemAttachments/ListItemAttachments.tsx
@@ -22,9 +22,13 @@ import { Spinner, SpinnerSize } from 'office-ui-fabric-react/lib/Spinner';
 import utilities from './utilities';
 import { Placeholder } from "../placeholder";
 
+interface IPreviewImageCollection {
+  [fileName: string]: IDocumentCardPreviewImage;
+}
+
 export class ListItemAttachments extends React.Component<IListItemAttachmentsProps, IListItemAttachmentsState> {
   private _spservice: SPservice;
-  private previewImages: IDocumentCardPreviewImage[];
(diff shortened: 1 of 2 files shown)


estruyf:
Thanks @ltdu - this will be released as a bug fix. As you mention that there wasn't working something before, it might be that there wasn't a lot of usages yet + we will bump the major version soon as we plan to update to Office UI Fabric version 6 + SPFx 1.8.2 or higher.

PR #312, opened 2019-05-23
Completed and fixed french translations, reordered based on mystrings…

Zhephyr54:
Updated FR translations and restructured localization file to follow mystring.d.ts file like #285.


estruyf:
@Zhephyr54 thank you for the update, this will be available in version `1.14.0`.

PR #317, opened 2019-05-29
Added ExtensionContext as a type for the context property

statto1974:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [* ]
| New sample?      | [ ]

Added ExtensionContext to the context property of all props. This allows for the control to be used in Extensions


AJIXuMuK:
Hi @statto1974,
I would discuss this one.
Maybe it's better to create `IContext` interface in the repo that will contain the properties that are needed for our controls. Because Extensions context and Web Part context are actually inherited from the same base class.
So i't much better if we have single property that can be used both in extensions and web parts. And we can expose this property to other components later on, not only for Taxonomy Picker.

@estruyf what do you think?

AJIXuMuK:
One more thing -  it looks like you're trying to create a PR from `master` branch` to `dev` which is not good. It's much better if you create a new branch based on `dev` and create a PR from `statto1974:your-new-branch` to `SharePoint:dev`

statto1974:
@AJIXuMuK This is my first PR, so it's all new to me 😁 Will create a fork from the dev branch

Agree on IContext idea, haven't looked through the other controls yet, but the TaxonomyPicker is only using spHttpClient and pageContext.web.absoluteUrl from the context passed in. This will stop them being tied to particular contexts and future proof for any new SPFx contexts that come along in the future.

estruyf:
In that case, I would hold it to the next major version release, as this would impact the already implemented controls.  That way, it would be better to just specify the properties we need, instead of passing the context.

statto1974:
@estruyf shall I close this PR? Happy to start looking at this for the next version

estruyf:
@statto1974 👍

PR #321, opened 2019-06-04
onClick and onDblClick should return void, not {}

AJIXuMuK:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | no

`onClick` and `onDoubleClick` properties were defined incorrectly in some `field` controls - the return type of these functions should be `void` instead of `{}`


estruyf:
Thanks @AJIXuMuK

PR #329, opened 2019-06-29
New control -  DataTrend 

joaojmendes:
| Q               | A
| --------------- | ---

| New feature?    | [X ]

Add DataTrend Control this is a implementation of react-trend component


estruyf:
Thanks @joaojmendes for this PR, but the control you added to this library is already a React control and does not provide any additional value to be integrated here. If you or any other developer want to use it, it is best to include the data trend library in your project directly which gives you a lot more options.

Will be closing this PR.

PR #338, opened 2019-07-19
DE localization

pfc2k8:
Updated German translations and restructured the localization file to follow mystrings.d.ts like #285 .
Also removed the de-de.ts file from the project root, that got accidentally committed by the previous editor of the German translations.

- DateTime picker translation is based on CLDR: [http://cldr.unicode.org/](http://cldr.unicode.org/)
- RTE translation is based on the OOTB SharePoint modern Text web part


estruyf:
@pfc2k8 thank you very much for adding and updating the German translations.

PR #341, opened 2019-07-23
Bug fix #150 (Taxonomy picker - Accepting AnchorId in typeahead search)

### LINKED ISSUE #150: TaxonomyPicker does not work correctly with anchorId
#### Category
[ ] Enhancement

[x] Bug

[ ] Question

#### Version 

Please specify what version of the library you are using: [ 1.9.0  ]

#### Observed Behavior
I use anchorId, to select terms from that level and below. In terms tree view it shows correct values. But auto complete ignores it and shows values from the whole tree… 

Thanks!

praveenbattula:
| Bug fix?        | [ x ]
AnchorId property is ignored in the typeahead search in taxonomy picker control. It works fine in the taxonomy picker panel. I added code to fix it by passing anchorId to service layer and filter terms by anchorId.

#### Guidance
- Accepting AnchorId (term node) in type ahead search in taxonomy picker.


praveenbattula:
Adding another pull request due to clean implementation instead of quick fix. So closing this one.

PR #343, opened 2019-07-24
#342 Add support for loginName with ensureUser.

### LINKED ISSUE #342: PeoplePicker: get loginName with ensureUser
#### Category
[x] Enhancement

[ ] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [1.14.0]

#### Observed Behavior
When 'ensureUser' parameter is set to true, selectedItems handler provides the sharepoint site user Id only. It is not supported to retrieve loginName and sharepoint user id at the same time.

Both properties: loginName and sharepoint id of the user are available.

siata13:
| Q               | A
| --------------- | ---
| Related issues?  | fixes #342

Add support for both loginName and sharepoint user id when ensureUser is set to true.

### DIFF
--- a/src/controls/peoplepicker/IPeoplePicker.ts
+++ b/src/controls/peoplepicker/IPeoplePicker.ts
@@ -108,7 +108,14 @@ export interface IPeoplePickerState {
 }
 
 export interface IPeoplePickerUserItem {
+  /**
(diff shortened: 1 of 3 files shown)


estruyf:
@siata13 thank you for this fix/enhancement.

PR #350, opened 2019-07-30
Initial version of the GridLayout control

hugoabernier:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [x]
| New sample?      | [ ]
| Related issues?  | 
| Related pull request | #351

#### What's in this Pull Request?

A GridLayout control to simplify rendering web parts with the standard grid layout. Intended to be really easy to use and handles reflowing content and compact layout.

Sample Grid Layout

Sample reflowing

PR #357, opened 2019-08-18
Filmstrip layout

hugoabernier:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [x]
| New sample?      | [ ]
| Related issues?  | 
| Related pull request | #350

A filmstrip control to simplify rendering web parts with the standard filmstrip layout. 

Sample Filmstrip Layout


estruyf:
@hugoabernier recently a carousel control was added. Might be better to enhance it with a couple of things you have implemented otherwise we have two similar controls.

hugoabernier:
@estruyf Good point. I was trying to create a set of controls that would allow developers to create web parts that follow the [SharePoint Design](https://docs.microsoft.com/en-us/sharepoint/dev/design/layout-patterns) principles.

My goal was to provide one "layout" control for each of the standard SharePoint layouts:
image

The `filmstripLayout` (and `carouselLayout`) are designed to behave exactly like the standard SharePoint layouts, complete with touch support, etc. They do not provide as much granular control as the existing `carousel` control (which is wonderful, by the way) because the goal was to replicate, not to improve upon, the standard layouts. As such, they serve different purposes.

I can see how having a `filmstripLayout`, a `carouselLayout`, and a `carousel` control would be confusing.

Would it be better to combine the `gridLayout`, `filmstripLayout`, `carouselLayout` and (not yet submitted) `compactLayout` into a single `layout` control?

Let me know and I'll gladly oblige.

PR #399, opened 2019-10-26
Fix for issue #398

### LINKED ISSUE #398: Placeholder: icon and text size are smaller than expected
#### Category
[ ] Enhancement

[x] Bug

[ ] Question

#### Version
Please specify what version of the library you are using: [ 1.15.0 ]

Here's the expected view of the Placeholder control:
image

#### Observed Behavior
- I add the Placeholder to my web part.
- Build and deploy a package.
- Icon and text are displayed much smaller than expected.
image

The behavior is the same for the workbench page and regular SharePoint Online pages.

#### Steps to Reproduce
Add the Placeholder control to a web part and run the project.

dmitryrogozhny:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #398

Issue: the Placeholder control is using global class names from Office UI Fabric. When no `fabric.css` file is available (I believe, this is the current default behavior in SharePoint Online), these classes do not work. Icon and text sizes will be smaller than expected in this case.

This PR changes global CSS class names with corresponding mixins, thus the Placeholder control does not depend on Office UI Fabric in runtime.


dmitryrogozhny:
Modified failing tests for the Placeholder:
- changed selector for the Icon component as the i tag won't have `ms-Icon--Add` class (data-icon-name attribute is set instead).
- renamed duplicate test 'Test placeholder without button' to 'Test that placeholder renders' as the test only checks that component renders.
- disabled a warning message for the Icon component - WARN: 'The icon "Add" was used but not registered. See http://aka.ms/fabric-icon-usage for more information.'
- changed test condition for 'Test placeholder with custom classname' - it should have a top-level <div> container with both standard and custom classes applied.

AJIXuMuK:
Thank you @dmitryrogozhny for the fix!

PR #404, opened 2019-10-29
Update FilePicker example props

danidz96:
Update FilePicker example accepts prop value with array of strings instead of string


AJIXuMuK:
Hi @danidz96!
Thanks for the fix!
It has been manually merged in [this commit](https://github.com/SharePoint/sp-dev-fx-controls-react/commit/8d86fcb890fac41e2e8c296b00de43f7183c7cb3)

PR #409, opened 2019-11-05
fix #405, when using drop down for time display, value is now formatt…

### LINKED ISSUE #405: DateTimePicker: dropdown for time not handling AM/PM correctly
[ ] Enhancement

[X ] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [   1.15.0      ]

Using the DateTimePicker with dropdown for time selection, AM and PM should be handled correctly.

#### Observed Behavior
When using TimeDisplayControlType.Dropdown, and choosing an hour value that is PM, when data is submitted to SharePoint Date/Time field, it is set as AM. Upon debugging the code, it looks like the issue is happening in the dropDownHoursChanged function. When the control loads, the value passed in is correct, where there is a space between the hours and AM or PM:
DateTimePickerInitial

However, on subsequent changes of the hour value in the drop down, the value comes in with no spaces and causes the following logic to break when trying to split the hour portion and designator portion on " "
DateTime4

DateTimePicker5

The hoursSplit array only has one value, "5PM" since the value is passed in without a space ("5PM"). The addition of 12 hours never happens since position 1 of the array is being looked at and the array does not have 2 items:
DateTimePicker2

I cannot find how/why the value comes in as expected on the initial render of the component (10 PM instead of 10PM for instance), but can see why the values are loaded without spaces for subsequent changes. The HoursComponent is storing the 1PM, 2PM, 3PM etc in the text property of the choice objects for the drop down. I'd be happy to take a shot at fixing the issue, but need to understand how the initial value is formatted properly.

This is possibly related to #402 but was not positive so created a new issue.

#### Steps to Reproduce
Add the DateTimePicker control with TimeDisplayControlType.Dropdown and choose an hour value in PM, then update a list item with a date/time field, using the value returned from the DateTimePicker control. It will set the time portion incorrectly.

c-eiser13:
…ed properly with space between hours and designator.

| Q               | A
| --------------- | ---
| Bug fix?        | [X ]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #405 

In the HoursComponent of the DateTimePicker, updated the onChange function of the Dropdown control to pass the value to the TimeHelper.hoursValue function so it will be formatted as the DateTimePicker component expects it to be. 

Previous: onChange(option.text)
New: onChange(TimeHelper.hoursValue(option.key as number, this.props.timeConvention))


AJIXuMuK:
Thanks for the fix @c-eiser13!

PR #422, opened 2019-11-30
valueColumnInternalName should be keyColumnInternalName

siddharth-vaghasia:
There seems to typo(mistake) in the sample code shown.... 
valueColumnInternalName does not exist as attribute in this conrol, it should be keyColumnInternalName.

Updated same in md file...

| Q               | A
| --------------- | ---
| Documentation fix      

Updated md file to correct attribute name as keyColumnInternalName instead of valueColumnInternalName in sample code.


AJIXuMuK:
Thanks for the fix @siddharth-vaghasia

PR #431, opened 2019-12-06
Fix doubled scroll issue when iframe content is higher than frame height

bbronisz:
| Q               | A
| --------------- | ---
| Bug fix?        | [ x]
| New feature?    | [ ]
| New sample?      | [ ]

It fixes double scrolls on different browsers due to wrong height calculations of iframe content, due to:
- selectors without . for classes
- marginTop style's value used to calculate marginBottom
- usage of style property on element (it was empty e.g. on Chrome) instead of window.getComputedStyle
- too small value used for padding (at least on Firefox and Chrome)


AJIXuMuK:
Great job @bbronisz!
Thank you for the fixes!

PR #437, opened 2019-12-13
Extended TaxonomyPicker with the ability to hide deprecated terms "Available for tagging" = false

### LINKED ISSUE #421: TaxonomyPicker: Possibility to hide deprecated and "Available for Tagging"=false term
#### Category
- [x] Enhancement
- [ ] Bug
- [ ] Question

#### Version
1.15.0

#### Desired Behavior
I would like to be able to configure the TaxonomyPicker to hide terms that are not selectable by the user anyway, as for example "Available for Taggin"=false and deprecated terms. Please see below print screen. I guess in certain situations it might make sense to show these terms and in certain situations it might not make sense.

#### Observed Behavior
image
NYC - "Available for Tagging"=false
BER - Deprecated

JonasBjerke89:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [x]
| New sample?      | [ ]
| Related issues?  | fixes #421 

Two new properties on the ITaxonomyPicker interface that gives the consumer the ability to hide deprecated terms (instead of deactivation only) as well as "Available for tagging" = false. The service provider is extended with the two props and the terms are filtered out if used.

### DIFF
--- a/src/controls/taxonomyPicker/ITaxonomyPicker.ts
+++ b/src/controls/taxonomyPicker/ITaxonomyPicker.ts
@@ -65,6 +65,16 @@ export interface ITaxonomyPickerProps  {
   termActions?: ITermActions;
 
   /**
+   * Specifies if the tags marked with 'Available for tagging' = false should be hidden
+   */
+  hideTagsNotAvailableForTagging?: boolean;
+
+  /**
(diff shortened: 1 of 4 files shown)


AJIXuMuK:
Thanks for that enhancement @JonasBjerke89!

PR #460, opened 2020-01-14
#458 listPicker always return "test" when multiple allowed

### LINKED ISSUE #458: listPicker always return "test" when multiple allowed
#### Category
[ ] Enhancement

[x] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [ 1.16.0 ]

List picker should return the list of chosen or empty array or null when multiple is allowed

#### Observed Behavior
"test is always included as the first element from return even no item is chosen

#### Steps to Reproduce
It only happens when **multiSelect** is enabled
```
<ListPicker
                    context={this.props.context}
                    multiSelect={true}
                    label='Select your list(s)'
                    includeHidden={false}
                    onSelectionChanged={(selected) => console.log(selected)} />
```

Thanks!

mgwojciech:
| Q               | A
| --------------- | ---
| Bug fix?        | [ *]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #458 

Removed "test" from initial selection inListPicker.onChanged

### DIFF
--- a/src/controls/listPicker/ListPicker.tsx
+++ b/src/controls/listPicker/ListPicker.tsx
@@ -122,7 +122,7 @@ export class ListPicker extends React.Component<IListPickerProps, IListPickerSta
 
     if (multiSelect === true) {
       // Check if option was selected
-      let selectedLists = this._selectedList ? cloneDeep(this._selectedList) as string[] : ["test"];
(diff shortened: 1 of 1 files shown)


AJIXuMuK:
Thanks @mgwojciech for that fix! It'll be included in the next version. But you can test it in beta version right now.

PR #485, opened 2020-02-23
Icon picker with localisation and documentation

Konradox:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [x]
| New sample?      | [ ]
| Related issues?  | 

This pull request adds new control - IconPicker. This control allows to pick an icon from office-ui-fabric-react icon set.


AJIXuMuK:
Thank you @Konradox for the component!
I really like the idea to have an icon picker in the library.

The only thing I'm concerned - predefined list of icons' names. It will be hard to maintain if we change used office ui fabric version...

Maybe we could explicitly reference `@uifabric/icons` dependency (v 5.8.0 is the one used by `office-ui-fabric-react@5.131.0`) and use `IconNames` enum from there?
In that case we'll just need to update dependency, not maintain the list of names.

@estruyf - what are your thoughts?

Konradox:
You are right @AJIXuMuK. But it's not possible because this enum is a const enum:
`export declare const enum IconNames {
    PageLink = "PageLink",
    CommentSolid = "CommentSolid",`

This means that is not iterable. It's also marked as deprecated.

estruyf:
Maybe we need to find another way to get all the icon names. @juliemturner what was your approach to get all the available icon names?

Konradox:
In the newer version of @uifabric/icon there is a new folder called data and there is a file AllIconNames.json which contains all that we need. Maybe we could consider upgrading this package?

From @uifabric/icon github [changelog](https://github.com/OfficeDev/office-ui-fabric-react/blob/c22d1f067350cf21a877bdfbef0b0ff1f41a42c3/packages/icons/CHANGELOG.md):
7.1.0
Tue, 02 Jul 2019 12:29:06 GMT

Minor changes
Update MDL2 icons. Add AllIconNames.json

AJIXuMuK:
Hi @Konradox,

We can try to use v7.1.0 of @uifabric/icon and test if it works with current referenced version of OUIFR.
If not - we're considering updating components to v7 of OUIFR but it will happen after SPFx is completely updated to v7 as well.

russgove:
hey, this may help https://github.com/OfficeDev/office-ui-fabric-react/issues/11399

i upgraded the react-securitygrid sample webpart to show user selected icons based on a persons. permission level and needed the same thing.

Konradox:
@russgove that's the file I've linked and a solution that requires package update 😊

@AJIXuMuK I will check if the @uifabric/icon update will work.

AJIXuMuK:
Thank you @Konradox for checking this one.

@estruyf I think we can merge the control as is for now, but keep in mind that we'll need to update it after OUIFR update.

estruyf:
thanks @Konradox - merged this to our DEV branch

PR #504, opened 2020-03-19
Update Placeholder.md

hugoabernier:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [ ]
| New sample?      | [ ]
| Updated documentation      | [x]
| Related issues?  | 

Changed code sample so the `onConfigure` event handler does not require binding.


estruyf:
@hugoabernier thanks, but this is actually not a good practice. The binding should not happen in any render. As that would mean your component binds each time it renders over and over again. The best practice is to bind it in the constructor or use arrow functions on the methods like:

```TypeScript
private onConfigure = () => {
  ...
}
```

hugoabernier:
@estruyf thank you! I learned something new today. Can I suggest that we update the sample to add the constructor code then? I'll do a separate PR.

PR #514, opened 2020-03-26
Fix: additional check for OrgAssets HTTP request

gautamdsheth:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | NA

While using the FilePicker control, if I have set the `hideOrganisationalAssetTab` value to `true`, even then an additional HTTP request is made. 

This PR avoids that by making the HTTP request only when `hideOrganisationalAssetTab` value is set to `false`. Especially needed in case of OnPrem where we don't have this capability available.


AJIXuMuK:
Great contribution @gautamdsheth! Thank you!
The PR has been merged and the fix is available in beta version.

PR #521, opened 2020-03-30
Fix: icon picker search not working

### LINKED ISSUE #512: Search in IconPicker doesn't work at all
#### Category
[ ] Enhancement

[x ] Bug

[ ] Question

#### Version
[1.17.0]

Nothing happens when type text in search box.
Error in browser console: "IconPicker.js:52 Uncaught TypeError: Cannot read property 'length' of undefined"

Reason:
`onChange` handler expects 2 parameters (`_event` and `newValue`), but the method is called only with one parameter `newValue`.

Thanks!

gautamdsheth:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #512 

This PR fixes the issue with icon picker search.
Currently while searching for any icon, it shows the same default list of icons without filtering as per the search criteria.

### DIFF
--- a/src/controls/iconPicker/IconPicker.tsx
+++ b/src/controls/iconPicker/IconPicker.tsx
@@ -54,7 +54,8 @@ export class IconPicker extends React.Component<IIconPickerProps, IIconPickerSta
     private closePanel = (): void => {
         this.setState({
             currentIcon: null,
-            isPanelOpen: false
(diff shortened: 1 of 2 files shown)


AJIXuMuK:
You can also test it in beta version right away.

PR #527, opened 2020-04-01
Fix #513 - IconPicker selection fix

### LINKED ISSUE #513: IconPicker shows selected icon only during the first opening
#### Category
[ ] Enhancement

[x ] Bug

[ ] Question

#### Version
 [1.17.0]

#### Steps to Reproduce

1. Use `IconPicker` control with set `currentIcon` property in your app.
2. Open icon picker - everything is ok, icon is selected
3. Close picker
4. Open the same picker again - **current icon is not selected**

Thanks!

gautamdsheth:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #513 

This PR fixes #513 - icon picker value doesn't remain selected. This PR fixes it by adding an additional state variable and storing the value in that.

### DIFF
--- a/src/controls/iconPicker/IIconPickerState.ts
+++ b/src/controls/iconPicker/IIconPickerState.ts
@@ -2,4 +2,5 @@ export interface IIconPickerState {
     items: string[];
     currentIcon?: string;
     isPanelOpen: boolean;
+    selectedIcon?: string;
 }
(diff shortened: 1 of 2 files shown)


AJIXuMuK:
Hi @gautamdsheth!
Thank you for the changes. But there was an actual bug in the code. There is no need for additional property in the `state`.

PR #552, opened 2020-04-30
Add maxDate & minDate 

PrasadKasireddy:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [X]
| New sample?      | [ ]
| Related issues?  | fixes https://github.com/SharePoint/sp-dev-fx-controls-react/issues/497

To set DatePicker with date boundary (minDate, maxDate).

### DIFF
--- a/src/controls/dateTimePicker/DateTimePicker.tsx
+++ b/src/controls/dateTimePicker/DateTimePicker.tsx
@@ -15,6 +15,7 @@ import { Async, css } from 'office-ui-fabric-react/lib/Utilities';
 import { IDateTimePickerProps, IDateTimePickerState, DateTimePickerStrings } from ".";
 import { TimeHelper } from "./TimeHelper";
 import { TimeDisplayControlType } from "./TimeDisplayControlType";
+import { addMonths, addYears } from 'office-ui-fabric-react/lib/utilities/dateMath/DateMath';
 
 interface IDateComponents {
   day: Date;
@@ -53,7 +54,8 @@ export class DateTimePicker extends React.Component<IDateTimePickerProps, IDateT
       hours,
       minutes,
       seconds,
-      errorMessage: ''
+      errorMessage: '',
+      today: new Date(Date.now())
     };
 
     this.async = new Async(this);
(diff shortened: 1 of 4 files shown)


PrasadKasireddy:
@AJIXuMuK  - could you please review the PR

src/controls/dateTimePicker/IDateTimePickerState.ts
```diff
@@ -7,4 +7,5 @@ export interface IDateTimePickerState {
   minutes?: number;
   seconds?: number;
   errorMessage?: string;
+  today: Date;
```
AJIXuMuK, line 10: Why do we need `today` property in the state?
It's a never changing value.

src/controls/dateTimePicker/DateTimePicker.tsx
```diff
@@ -205,7 +207,9 @@ export class DateTimePicker extends React.Component<IDateTimePickerProps, IDateT
       strings: dateStrings = new DateTimePickerStrings(), // Defines the DatePicker control labels
       timeDisplayControlType,
       placeholder,
-      showLabels
+      showLabels,
+      minDate = addYears(this.state.today, 1),
+      maxDate = addYears(this.state.today, 1)
```
AJIXuMuK, line 212: I don't think we need to set some values to `minDate` and `maxDate` by default.
Moreover, here both values are set to the same value.

AJIXuMuK: I've made the chage

src/controls/dateTimePicker/IDateTimePickerState.ts
AJIXuMuK: I've made the change

AJIXuMuK:
Hi @PrasadKasireddy - I've made additional changes and merged the PR.

Thanks for the contribution!

PrasadKasireddy:
> Hi @PrasadKasireddy - I've made additional changes and merged the PR.
> 
> Thanks for the contribution!

@AJIXuMuK  Thanks

PR #571, opened 2020-05-26
Correct the AnchorID getting all TermSet search options

DRamalho92:
| Q          | A
| --------------- | ---
| Bug fix?        | [ X]

When using the Option AnchorID on the TaxonomyPicker, it only shows on the panel the correct option to pick from, but when using the type, it'll display all the other terms of the terms. 

AnchorID

#150 
#341


estruyf:
Thanks @DRamalho92 for the fix! It will be part of the `1.19.0` release.

PR #625, opened 2020-07-30
TaxonomyPicker: Include check for separator while filtering path of terms when anchorId is configured

devangbhavsar89:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]

Bug fix related to anchorId configuration inside TaxonomyPicker control - When anchorId is used with TaxonomyPicker control, and if there exists any other term (that is not as a child of anchor term and it starts with same name as the anchor term name), then that term becomes available in the terms listing. It should not be listed because it is not a child term.

**Bug Description**

Say, we have hierarchy of terms in the term store as follows:
image

Now if we configure TaxonomyPicker control to use "Helsinki" as anchorId, then TaxonomyPicker control shows results as follows:

1. While picking term from hierarchy listing
image

2. While filtering terms via. type-in
image

If we check the hierarchy of terms inside the term store, then "Helsinki, Espoo, Vantaa" is not a child term of "Helsinki". So ideally, it should not be listed/populated inside our control.

This PR fixes above mentioned issue of TaxonomyPicker control.

### DIFF
--- a/src/controls/taxonomyPicker/TermParent.tsx
+++ b/src/controls/taxonomyPicker/TermParent.tsx
@@ -37,9 +37,9 @@ export default class TermParent extends React.Component<ITermParentProps, ITermP
     {
       const anchorTerm = this._terms.filter(t => t.Id.toLowerCase() === this.props.anchorId.toLowerCase()).shift();
       if (anchorTerm) {
-        const anchorDepth = anchorTerm.PathDepth;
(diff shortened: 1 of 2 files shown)


estruyf:
Thanks @devangbhavsar89 - this is merged and will soon be available as a beta.

Issue #624, opened 2020-07-30
TaxonomyPicker Suggesstion not working in SharePoint Online with no error in console


desaiprerak:
#### Category
[ ] Bug

#### Expected / Desired Behavior / Question
Suggestions should load

#### Observed Behavior
Suggesstion not loading

#### Error Details
Network reponse
ErrorCode: -1
ErrorMessage: "Guid should contain 32 digits with 4 dashes (xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx)."
ErrorTypeName: "System.FormatException"
ErrorValue: null

Thanks!
Prerak

jkib:
I am having the same issue - I've spent the last day or so thinking I must have implemented the control incorrectly. Glad I'm not alone.

desaiprerak:
Is there any update on issue

desaiprerak:
> #### Category
> [ ] Bug
> 
> #### Expected / Desired Behavior / Question
> Suggestions should load
> 
> #### Observed Behavior
> Suggesstion not loading
> 
> ####Error Details
> ErrorCode: -1
> ErrorMessage: "Guid should contain 32 digits with 4 dashes (xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx)."
> ErrorTypeName: "System.FormatException"
> ErrorValue: null
> 
> Thanks!
> Prerak

jkib:
Desaiprerak, my suggestions started working on Friday - though to be honest I have no idea why because I didn't make any changes to fix it. I delivered it to a user, told them that the suggestions didn't work yet, and they reported that it did work..

TheThor59:
Hello, correction is currently on the way.
I had the same issue and found bug #629 (created more recently). I tested on the dev branch and a fix has been made so it will come on next version

jkib:
Thanks!

AJIXuMuK:
Thank you @desaiprerak for reporting this.
It seems like a duplicate of #604 that has been fixed and will be available in the next version.

Issue #604, opened 2020-07-08
Pnp TaxonomyPicker suggestion/match does not work as expected


geltapatio:
#### Category
[ ] Enhancement

[X ] Bug

[ ] Question

#### Expected / Desired Behavior / Question
you can find any term, which is part of the configurated taxonomy hierarchy 
if the user searches for a subordinate term suggestion/match will display it.

#### Observed Behavior
The suggestion/match of TaxonomyPicker only finds main terms. The subterms could not be found.

#### Steps to Reproduce

Example 

if you search with "Kap" suggestion/match will display "Kapitel1"
if you search with "Th" suggestion/match will not display results
if you search with "Ab" suggestion/match will not display results
[] Kapitel1
--- [] Thema1
--- --- [] Ablauf1
--- --- [] Ablauf2
--- --- [] Ablauf3
--- --- [] Ablauf4

![bidl1_tems](https://user-images.githubusercontent.com/9026810/87123573-56272180-c287-11ea-95dd-4474bcab3ee8.JPG)

Current Version

Main Term (Kapitel1) was found
![bild2_KapitelOK](https://user-images.githubusercontent.com/9026810/87123574-56272180-c287-11ea-9093-5f1051820298.JPG)

subordinates (Ablauf1, Ablauf1_3ToDelete, Ablauf2, Ablauf3, Ablauf4) could not be found.
![bild3_AblaufNOK](https://user-images.githubusercontent.com/9026810/87123575-56bfb800-c287-11ea-8e74-5ebff484bac8.JPG)

subordinate (Thema1) could not be found
![bild3_ThemaNOK](https://user-images.githubusercontent.com/9026810/87123576-56bfb800-c287-11ea-882a-e3cb3434d255.JPG)

Thanks!
Luis Miguel Gamiz Espinosa

geltapatio:
I already fixed it and created a pull request

AJIXuMuK:
@geltapatio - thank you for reporting and fixing the issues.

The PR has been merged and will be available in the next release.

In a meanwhile please test the functionality in [beta version](https://pnp.github.io/sp-dev-fx-controls-react/beta/).

TheThor59:
Hello @AJIXuMuK 

could we mark #624 and #629 as duplicate of this ticket as a fix has been impremented on beta ?
Alors I've just made a pull request to fix the new correction as there is an issue when using empty GUID, this is linked to this point.

Thanks

limcysam:
[...]

Issue #288, opened 2019-04-30
TaxonomyPicker: finding terms with label


JarecN:
#### Category
[X] Enhancement

[ ] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [ 1.12.0 ]

_If you are not using the latest release, please update and see if the issue is resolved before submitting an issue._

#### Expected / Desired Behavior / Question
As term could have other labels (synonyms and abbreviations) it would be great to be able to find a term by typing other label, not just typing default label. Let's say I have term which default label is "Internal Memo" and it has other labels "Executive Board Memo" and "HR Memo". Now when I type "executive" or "hr" on TaxonomyPicker I don't get any suggestions. I would love to get that "Internal Memo" suggested when typing "executive" or "hr".

Thanks!

TheThor59:
Hello,

I've implemented a version that should do the trick for alternate labels on terms. I didn't had solution for alternative languages. If any one has solution to detect user's current LCID (1033) and not local (en-US) that could be perfect to fine tune the feature.
The commit somehow went into my pull request #631 so it complete the fix I did here on issue with empty anchorId

PR #631, opened 2020-08-12
Added correction that taxonomy picker crashed when SharePoint return …

### LINKED ISSUE #629: TaxonomyPicker suggestion not working in SharePoint Online
#### Category

[ ] Bug

#### Version

Please specify what version of the library you are using: [ 1.19.0  ]

Suggestion load as when clicked on the icon

#### Observed Behavior
Suggestion do not load with spinning wheel in type ahead

#### Steps to Reproduce
Type in a existing term within the textbox

TaxonomyPickerIssue

TheThor59:
Q | A
-- | --
Bug fix? | [ X ]
New feature? | [ ]
New sample? | [ ]
Related issues? | fixes #629, #288, #624, #604  

Following ticket #629, I found that a fix has been made on DEV branch. I tested but an issue appear when providing to parameter "AnchorId" the value that return SharePoint on loading list fields. The default response of SharePoint is empty GUID and not undefined, making the suggestion system crash. I added a check on that point to handle the case

Regards.

### DIFF
--- a/src/common/Constants.ts
+++ b/src/common/Constants.ts
@@ -109,3 +109,5 @@ export const InvariantCulture: ICultureInfo = {
         eras: [1, 'A.D.', null, 0]
     }
 };
+
+export const EmptyGuid: string = "00000000-0000-0000-0000-000000000000";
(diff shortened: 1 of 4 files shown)


AJIXuMuK:
Thanks @TheThor59 for these changes!

PR #638, opened 2020-08-21
Add Accordion control from https://github.com/hugoabernier/AccordionWebPart

sadomovalex:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [x ]
| New sample?      | [ ]

Add Accordion control from https://github.com/hugoabernier/AccordionWebPart

### DIFF
--- a/src/Accordion.ts
+++ b/src/Accordion.ts
@@ -1,0 +1,1 @@
+export * from "./controls/accordion/index";
--- a/src/controls/accordion/Accordion.module.scss
+++ b/src/controls/accordion/Accordion.module.scss
@@ -1,0 +1,49 @@
+@import "~office-ui-fabric-react/dist/sass/References.scss";
+
+.accordion {
+  background: #fafafa;
+
+  & > button {
+    min-height: 44px;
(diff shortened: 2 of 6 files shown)


estruyf:
@AJIXuMuK it is because of the dependency of Office UI Fabric 6.

AJIXuMuK:
@estruyf do you see the reason for that? 
I looked at the code and it seems like it could be based on v5.

estruyf:
Indeed, seems just because the usage of `DefaultButton`.

estruyf:
If we replace that, we can make it happen in `v1`.

estruyf:
ok, seems that it was just a couple of properties that have been renamed. I made changes so that it works fine in v1. Thanks, @sadomovalex for this PR.

sadomovalex:
Great, it is even better!

sadomovalex:
@AJIXuMuK I'm not author of this control. Your thanks should go to @hugoabernier :) I just wanted to make it available with less efforts directly from npm)

PR #642, opened 2020-08-25
Update to FilePicker OrganisationAssetslibrary 

Timhaka:
| Q               | A
| --------------- | ---
| Bug fix?        | [X]
| New feature?    | [ ]
| New sample?      | [ ]

Small change to the OrgAssetsService.ts.
if your organisationAssets library dosent have a thumbnail image, it will crash and not display your organisation library as the 
libItem.ThumbnailUrl will be null. 

Made a quick change to see if i null check the Thumbnail instead if that would fix the problem witch it does.
Would require some additional testing. But in my environment i did not get any errors after the change.

Also thanks to Rabia Williams for finding where the error happened.


AJIXuMuK:
Hi @Timhaka!
Thank you for the PR!

This change has already been added and available in the [beta version](https://pnp.github.io/sp-dev-fx-controls-react/beta/)

PR #646, opened 2020-08-31
Fix/people search service profile picture image url

juhaalhojoki:
| Q               | A
| --------------- | ---
| Bug fix?        | [ x]
| New feature?    | [ ]
| New sample?      | [ ]

Noticed that an image for a user picked in PeoplePicker didn't get resolved on my dev tenant for my test users even when they have profile picture set.
PeopleSearchService: 
Issue: The generateUserPhotoLink(value: string) method that resolves user's profile picture URL takes user's email address as an input parameter as the parameter for the accountName query string parameter. If users don't have email address set or email address does not match user's account name, image cannot be resolved.
Fix: using Description field from the response solves the issue. Description field contains user's account name.


AJIXuMuK:
Thank you @juhaalhojoki!
Great contribution!

Issue #665, opened 2020-09-12
How to detect when the file picker is canceled?


TheInvoker:
#### Category
 Bug
#### Version

How can I detect when the dialog closes (canceled)?

AJIXuMuK:
Hi @TheInvoker,

Please, take a look at the latest version of the controls.
You should have `onCancel` event.

PR #668, opened 2020-09-16
Feature: added additional props for File Picker

gautamdsheth:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [x]
| New sample?      | [ ]
| Related issues?  | NA

Added 2 additional properties to the file picker control.

1) isPanelOpen - this will directly open the file picker panel
2) onCancel - this will provide a callback when the file picker is cancelled. This was mentioned in #665 

Can you also add this to the v1 branch as well ? Let me know if I need to open a separate PR for the same.


estruyf:
Thank you @gautamdsheth!

PR #687, opened 2020-10-07
Feature/fix - changed API for OrgAssets items

gautamdsheth:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [x]
| New sample?      | [ ]
| Related issues?  | not sure

Have changed the API used to fetch Org Assets items. It is the same as the one used by Microsoft now.
The current API requires very high level permissions(SharePoint admin)  to access items from org asset libraries.
Using the API mentioned in the PR, this will be permission trimmed to the current user and not require SharePoint admin privileges. Using this any user with permissions to the libraries will be able to view the items.

Have targeted v2 branch, please let me know if you also need the same PR in v1 branch as well. Will be happy to submit it again for that !

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "build": "gulp build",
     "clean": "gulp clean",
     "test": "gulp test",
+    "serve": "gulp serve",
     "prepublishOnly": "gulp",
     "versionUpdater": "gulp versionUpdater",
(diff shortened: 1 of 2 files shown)


AJIXuMuK:
Thanks @gautamdsheth for the update!

PR #688, opened 2020-10-07
Feature/fix - org assets API change for v1 branch

gautamdsheth:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [x]
| New sample?      | [x]
| Related issues?  | NA

Same as PR #687  , just changed the target branch to v1-dev


AJIXuMuK:
Thanks @gautamdsheth for adding this to `v1`.
But `v1` is applicable for SP on-prem only. So, we don't need that in `v1`.

Issue #677, opened 2020-10-05
TreeView control checkboxes checked state are not retained


GokulPT:
Hi,

I am facing a couple of issue in TreeView control in 1.20.0 version. 

1) When I expand a parent node, all the child nodes are expanded. I would like to know, if there is any property available to disable this behavior. The child nodes need to be in collapsed state and the user should expand them
2) The checkboxes checked state of the child nodes are not retained when we move from one screen to an another

Abderahman88:
For the first question, I submitted a PR.
For the second question, do you mean when you change tabs in your browser? I've tried this and the states of the child nodes didn't change. (PS: I tested with Google Chrome)

AJIXuMuK:
Hi @GokulPT!
As @Abderahman88 mentioned the first one has been implemented and will be available in the next release.
For the second one: if you **delete** the controls from the DOM while switching between screens then yes, the state will be lost as the controls will be re-created. It's a correct behavior for any React control.

So, if you have something like
```typescript
{treeViewScreen && <TreeView .../>}
{otherScreen && <OtherComponent .../>}
```
you will lose the state.
If you change it to something like:
```typescript
<div style={{ display: treeViewScreen ? 'block' : 'none' }}>
  <TreeView ... />
</div>
```
`TreeView` will remain the state.

PR #698, opened 2020-10-15
Treeview child nodes expand/collapse

Abderahman88:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [X]
| New sample?      | [ ]
| Related issues?  | mentioned in #677 

**Question** : When I expand a parent node, all the child nodes are expanded. I would like to know, if there is any property available to disable this behavior. The child nodes need to be in collapsed state and the user should expand them

**Solution** : Added (optional) property 'defaultExpandedChildren' that controls the behavior of the expansion of child elements.
if defaultExpandedChildren is set to false --> only the parent elements will be shown. (img. 1)
if defaultExpandedChildren is set to true --> all underlying child elements will be shown. (img. 2)

d1c0

d1c1

### DIFF
--- a/src/controls/treeView/ITreeViewProps.ts
+++ b/src/controls/treeView/ITreeViewProps.ts
@@ -67,4 +67,9 @@ export interface ITreeViewProps {
    * @argument item The tree item.
    */
   onRenderItem?: (item: ITreeItem) => JSX.Element;
+   /**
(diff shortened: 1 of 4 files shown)


AJIXuMuK:
Thank you @Abderahman88 for the enhancement

PR #729, opened 2020-11-11
TaxonomyPicker: Show an error message for an invalid/unresolved input

### LINKED ISSUE #728: TaxonomyPicker control does not show an error message for an invalid/unresolved input
#### Category
- [x] Enhancement
- [ ] Bug
- [ ] Question

#### Version

Version of the library: [2.1.0]

I recently came across a scenario, where it would be beneficial to indicate some sort of an error beneath the control whenever user inputs any invalid text (which is not able to resolve into a valid term) inside taxonomy picker control. 

The same kind of behavior is also visible with **OOTB Page Details** action (that opens a panel on right side). 
image

Implementation Suggestion: It can be achieved inside **`<TaxonomyPicker />`** control by triggering **onBlur()** and **onInputChange()** handlers that exists in **TermBasePicker** base component. All necessary validations can be done inside these handlers and an error message can be shown accordingly.

#### Observed Behavior
At the moment, such invalid/unresolved input is just ignored. And hence, it goes unnoticed from end-users point of view, while submitting the form.

Thanks!

devangbhavsar89:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| Enhancement?    | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #728

Provides ability to optionally show an error message beneath the taxonomy picker control for any invalid input entered by the user. 

#### Implementation Details
The input validation will happen based on newly introduced property named **"validateInput"** of type boolean. 
- If set to **true**, then an error message **[Please fix invalid term(s)]** will be shown for any invalid input. 
- If set to **false** or **not defined**, then it will continue to work as it works currently (i.e., no error message will be displayed).

#### Usage 
In order to enable input text validation, you need to set **`validateInput={true}`** property of **`<TaxonomyPicker />`** control.

image

### DIFF
--- a/src/controls/taxonomyPicker/ITaxonomyPicker.ts
+++ b/src/controls/taxonomyPicker/ITaxonomyPicker.ts
@@ -85,6 +85,11 @@ export interface ITaxonomyPickerProps  {
   validateOnLoad?: boolean;
 
   /**
+   * Specifies if the input text will be validated, when the component focus is changed
+   */
+  validateInput?: boolean;
+
+  /**
    * The method is used to get the validation error message and determine whether the input value is valid or not.
    * Mutually exclusive with the static string errorMessage (it will take precedence over this).
    *
--- a/src/controls/taxonomyPicker/TaxonomyPicker.tsx
+++ b/src/controls/taxonomyPicker/TaxonomyPicker.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { PrimaryButton, DefaultButton, IconButton } from 'office-ui-fabric-react/lib/Button';
 import { Panel, PanelType } from 'office-ui-fabric-react/lib/Panel';
+import { Autofill } from 'office-ui-fabric-react/lib/components/Autofill/Autofill';
 import { Spinner, SpinnerType } from 'office-ui-fabric-react/lib/Spinner';
 import { Label } from 'office-ui-fabric-react/lib/Label';
 import TermPicker from './TermPicker';
@@ -32,6 +33,7 @@ export const TERM_IMG = 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABAAAAAQC
 export class TaxonomyPicker extends React.Component<ITaxonomyPickerProps, ITaxonomyPickerState> {
   private termsService: SPTermStorePickerService;
   private previousValues: IPickerTerms = [];
+  private invalidTerm: string = null;
   private cancel: boolean = true;
 
   /**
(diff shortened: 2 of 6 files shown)


AJIXuMuK:
Hi @devangbhavsar89,
Thanks for the PR.

I have a question though. How's the new property behave (or should behave) together with `onGetErrorMessage` prop set?

devangbhavsar89:
Hi @AJIXuMuK, 

This  new text validation via. `validateInput` prop works independently from `onGetErrorMessage` prop.

Here is the output, with different set of configurations around those two props:

#### Scenario-1:
We set both `validateInput` and `onGetErrorMessage` props.

`<TaxonomyPicker ... validateInput={true} ... onGetErrorMessage={(value) => { return !!value && value.length > 0 ? '' : 'This field is required'; }} />`

Then the validation will be executed with following logic:
- Initial state
image

- When I add invalid term inside the text box of taxonomy picker control, **an error indicating invalid term will be displayed**.
image

- When I remove all terms from the taxonomy picker control
image

#### Scenario-2:
We only set `onGetErrorMessage` props. That is `validateInput` is either not configured or set to `false`.

`<TaxonomyPicker ... onGetErrorMessage={(value) => { return !!value && value.length > 0 ? '' : 'This field is required'; }} />` or `<TaxonomyPicker ... validateInput={false} ... onGetErrorMessage={(value) => { return !!value && value.length > 0 ? '' : 'This field is required'; }} />`

Then the validation will be executed with following logic:
- Initial state
image

- When I add invalid term inside the text box of taxonomy picker control, **no error related to invalid term will be displayed**.
image

- When I remove all terms from the taxonomy picker control
image

So basically the idea of using that additional `validateInput` prop, is to not force that 'Please fix invalid term(s)' error message each time (as this may not be the desired behaviour for all use-cases). Hence I've kept this additional validation around unresolved input text as optional (which is driven by `validateInput` prop).

Hope this answers  your question. Please let me know, if you need more details around it.

Thanks!

AJIXuMuK:
Thank you @devangbhavsar89 for the detailed explanation!
I've merged the PR.

Matej4386:
Hello @devangbhavsar89,
this is great ! It is possible to implement this PR also for v1 realease?

devangbhavsar89:
Hi @Matej4386,

I just checked the code base of v1 and it seems pretty much doable. So I'll implement this enhancement also for `v1` release and create a PR for it.

Matej4386:
@devangbhavsar89 thank you very much. I really appreciate it 👍

devangbhavsar89:
Hi guys!

@Matej4386 - I've created a PR #738, which includes same enhancement for `v1` release.

@AJIXuMuK - I noticed that we do have a milestone [1.21.1](https://github.com/pnp/sp-dev-fx-controls-react/milestone/25) for `v1`. So hopefully, this will be included in `v1` release.

AJIXuMuK:
Hi @devangbhavsar89, @Matej4386,

I've merged the PR, so it will be included in `v1.21.1`

PR #745, opened 2020-11-26
Better handling of session storage values

gautamdsheth:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | NA

Sometimes due to large number of items in the session storage, we get error about quota exceeded. This happened once due to large number of items in the Taxonomy picker as well which caused the picker not work. 
So, added a try catch block to prevent this.

Credit to @poikjo for finding the issue.


AJIXuMuK:

PR #759, opened 2020-12-08
Feature - Added session storage as property

gautamdsheth:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [x]
| New sample?      | [ ]
| Related issues?  | NA

Added `useSessionStorage` property to Taxonomy picker control.
Needed to add it as sometimes we need fresh data after doing some CRUD operations on the terms. 
Default will be set to true, but if set to false it will always fetch data from the term store.


AJIXuMuK:
Thank you @gautamdsheth for the enhancement!

PR #762, opened 2020-12-11
Fix - Return TermSetId for suggestions

gautamdsheth:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | NA

In some cases, there is a weird behavior in the Taxonomy picker control.

If you select the term from the panel, it returns the termSetId of the selected Term.
If you use suggestions, then it doesn't happen but we get default Empty GUID. This causes some issues downstream if you want to use termSetId further.
This PR fixes that by sending back the termSetId in case of suggestions as well.


AJIXuMuK:

Issue #765, opened 2020-12-14
Taxonomy Picker Control : Select Child items automatically


mahinthan:
#### Category
[ ] Enhancement

[ ] Bug

[ x ] Question

#### Version

Please specify what version of the library you are using: [   1.20.0     ]

#### Question
I'm using taxonomy picker control. When terms are heirarchical, while selecting the parent term, its not automatically selecting the child terms underneath that parent term.
I read AnchorId to be used to achieve. Do you have any working sample/etc to show how it can be used?

#### Observed Behavior
![image](https://user-images.githubusercontent.com/8817638/102063872-e1887e00-3df6-11eb-84f5-3d2f94137b82.png)

AJIXuMuK:
Thank you @mahinthan for providing the feedback!

I've implemented the change - you can use `selectChildrenIfParentSelected` property of the `TaxonomyPicker`.
The change will be included in the next release.

In a meanwhile you can test the functionality [beta version](https://pnp.github.io/sp-dev-fx-controls-react/beta/)

mahinthan:
Thanks @AJIXuMuK

PR #774, opened 2020-12-27
Fix duplicate when higher sub Term is selected "selectChildrenIfParentSelected".

aaclage:
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | partially #765, mentioned in #765 

I use the property "selectChildrenIfParentSelected" to test the new functionality and think to found a issue when selecting terms on following sequence.
If selected Term(for the example Switzerland), selects all sub terms associated what is expected, but if select "countries" the terms are added twice, plus "countries".
This code fix validate existing and don't include twice.

image

image

### DIFF
--- a/src/controls/taxonomyPicker/TaxonomyPicker.tsx
+++ b/src/controls/taxonomyPicker/TaxonomyPicker.tsx
@@ -283,14 +283,16 @@ export class TaxonomyPicker extends React.Component<ITaxonomyPickerProps, ITaxon
       }
 
       if (children.length) {
-        activeNodes.push(...children.map(c => {
-          return {
-            name: c.Name,
-            key: c.Id,
-            path: c.PathOfTerm,
-            termSet: c.TermSet.Id
-          };
-        }));
+        [...children].map(c => {
+          if (activeNodes.find(item => item.key == c.Id) === undefined) {
+            activeNodes.push({
+              name: c.Name,
+              key: c.Id,
+              path: c.PathOfTerm,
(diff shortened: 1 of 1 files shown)


aaclage:
Another issue that i find and wanted to validate with you @AJIXuMuK, want to know if this error also appears to you or only to me.
When i select a node the children's items are not selected and found the reason. It was in the code with operator (!) that should work as expected, since this is JS should work fine but it's not working (maybe some wrong mapping from typescript).
image
**Term.tsx**
Doesn't work (Tested in Chrome and Edge)
```
  public componentWillReceiveProps?(nextProps: ITermProps, nextContext: any): void {
    if (!this.props.multiSelection ) {
```
changed code below work as expected (Tested in Chrome and Edge)
```
  public componentWillReceiveProps?(nextProps: ITermProps, nextContext: any): void {
    if (this.props.multiSelection !== false) {
```
**PS:** didn't include in the code without your validation

AJIXuMuK:
Hi @aaclage!
Thanks for your tests and contribution!

For the duplicates - there already was a code to exclude duplicates. But I didn't move it to correct line while implementing the feature. So, I've just modified the initial code instead of merging the PR. 
I will still include you in the list of contributors though.

For the second issue - honestly, I don't think this condition should be there at all. I removed it and tested all possible scenarios of multiple selections + select children with parent.
Now everything looks good.

Again, thank you for the tests and contribution!

PR #789, opened 2021-01-24
ListView: Filter functions from props when checking equality

### LINKED ISSUE #251: ListView select not working
#### Category
- [ ] Enhancement
- [x] Bug
- [ ] Question

#### Version
1.12.0

When putting a ListView inside a React Component that controls its items and selection props, an item should be selectable.

#### Observed Behavior
When putting a ListView inside a React Component, anything that trigger componentDidUpdate, like setState or a prop change, will cause an invalidation of the items selection and cause an invalid call to props.selection().

#### Steps to Reproduce
Call setState on a component containing a ListView with a selected item and watch the selection disapear.
Handy reproduction: https://github.com/thespooler/listview-bughunt

thespooler:
| Q               | A
| --------------- | ---
| Bug fix?        | [ x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #251

This fixes the issue with #251 by filtering known functions from isEqual comparison. It might create new issues when selection or render function changes, but comparing functions for equality is a near impossible job.

### DIFF
--- a/src/controls/listView/ListView.tsx
+++ b/src/controls/listView/ListView.tsx
@@ -111,7 +111,7 @@ export class ListView extends React.Component<IListViewProps, IListViewState> {
    */
   public componentDidUpdate(prevProps: IListViewProps, prevState: IListViewState): void {
 
-    if (!isEqual(prevProps, this.props)) {
+    if (!isEqual(this._filterFunctions(prevProps), this._filterFunctions(this.props))) {
       // select default items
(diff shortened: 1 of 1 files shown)


AJIXuMuK:
Thank you @thespooler for the fix!
I've merged the PR with some additional changes to omit any possible functions in `items` and any future functions in `IListViewProps`

PR #814, opened 2021-03-04
Fix case sensitive fluent icon search service

mbice:
| Q               | A
| --------------- | ---
| Bug fix?        | [X]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | Nothing logged I could find

There is an issue in the Icon Picker control where searching for Fluent icon names is case sensitive. For example, searching for "aadlogo" will return an icon, whereas searching for "AADLogo", which is the friendly name of the icon, returns nothing.

Searching for "aadlogo"

image

Searching for "AADLogo"

image

This fix ensures that the casing of the search term is ignored so that icons are returned regardless of lower/uppercase characters.

Thanks!

### DIFF
--- a/src/services/FluentIconsService.ts
+++ b/src/services/FluentIconsService.ts
@@ -20,7 +20,7 @@ export class FluentIconsService {
 
   public search = (query: string, startsWith?: boolean): string[] => {
     const lowerCasedQuery = query.toLowerCase();
-    return this._iconNames.filter(name => startsWith === false ? name.toLowerCase().indexOf(lowerCasedQuery) !== -1 : name.toLowerCase().indexOf(query) === 0);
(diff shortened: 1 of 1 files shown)


AJIXuMuK:
Thank you @mbice for the fix!

I've merged it with a small change: I reverted back `!startsWith` to `startsWith === false` because this change reverses the default behavior.

mbice:
Thanks for the merge, @AJIXuMuK.

Sorry, I should have documented the change to "startsWith" as well. I noticed that this method is not always being called with the startsWith parameter, as it's optional. In the case of the iconPicker, nothing gets passed in, so `startsWith === false` will never evaluate as true. In the case of the icon picker this means it will always string match from the starting of the string. I changed it to !startsWith so that either explicitly passed in as "false" or not provided would be true.

Thoughts?

Here's the call from icon picker:

image

AJIXuMuK:
Got it @mbice!

I added it back and also added a new prop to the controls to be able to switch the behavior.

PR #832, opened 2021-03-13
[Treeview] Fix two potential null reference issues

YannickRe:
| Q               | A
| --------------- | ---
| Bug fix?        | [X]

A fix for two potential null reference issues:
- when `expandToSelected == true` but `defaultSelectedKeys == null`. This combination of properties does not make much sense : if you want  `expandToSelected` you NEED values in defaultSelectedKeys, but `defaultSelectedKeys` is not required thus this combination can happen and it breaks the control.
- when `expandToSelected == true` but somewhere in the tree there is a `treeItem.children == null`. You should set `children == null` when you don't want the expand arrow to show, but it breaks in combination with `expandToSelection == true`.

This PR fixes both issues.


AJIXuMuK:
Thank you @YannickRe  for the PR!

PR #864, opened 2021-04-05
[V2] Fix default item selection in TreeView

YannickRe:
| Q               | A
| --------------- | ---
| Bug fix?        | [X]

Fixes the selection of children when passed in as `defaultSelectedKeys`. Before, when a parent matched a key the children weren't checked any more if they were also part of `defaultSelectedKeys`. As a result, only the parent showed as selected. This is now fixed.

Deprecated `selectChildrenIfParentSelected` and added `selectChildrenMode`. This is a flagged enum which allows much more flexibility on when this before should be triggered. Currently, the behavior is triggered on mount, update (when new props are passed), select and unselect. As a result, you can never pass in the key of a parent and just one of its children as all of the children will be selected. With the change to a flagged enum, you can control all 4 lifetime moments separately.

Old behavior can be replicated with `SelectChildrenMode.None` and `SelectChildrenMode.All`. For more granular control can also be set to eg. `SelectChildrenMode.Select | SelectChildrenMode.Unselect`, which means that children will be selected/unselected when the parent is selected or unselected but not when the control mounts OR when the props are updated.

### DIFF
--- a/docs/documentation/docs/controls/TreeView.md
+++ b/docs/documentation/docs/controls/TreeView.md
@@ -34,7 +34,7 @@ import { TreeView, ITreeItem, TreeViewSelectionMode } from "@pnp/spfx-controls-r
     items={treeitems}
     defaultExpanded={false}
     selectionMode={TreeViewSelectionMode.Multiple}
-    selectChildrenIfParentSelected={true}
+    selectChildrenMode={SelectChildrenMode.Select | SelectChildrenMode.Unselect}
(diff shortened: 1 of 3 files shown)


AJIXuMuK:
Thank you @YannickRe!
The PR has been merged and is available in `v2-beta` for tests.

AJIXuMuK:
I've also cherry-picked it to `v3`

PR #871, opened 2021-04-08
Fix #799 React error in FileBrowser paging

### LINKED ISSUE #799: PnP SPFx DateTimePicker Clear Date functionality
As a end I don't have feasibility to clear the selected date with the PnP SPFx DateTimePicker control. Could some one please help on this if this is already fixed

### LINKED ISSUE #826: FilePicker - React(?) crash on large folders
#### Category

- [ ] Enhancement
- [x] Bug
- [x] Question

#### Version

Please specify what version of the library you are using: [ 2.5.0 ]

Being able to scroll even in large folders

#### Observed Behavior
Tested on two sites with folders with lot of files ~15000 and about 35000 in one folder, in both cases whole react crashes and ui gets closed, `Objects are not valid as a React child (found: [object Promise])` is the message in debug console. "15000" files library does have only files with names in latin alphabet.

#### Steps to Reproduce
Create library with 15000+ files

I dunno if this can be solved, it is mostly a question if it is even possible.

cwparsons:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #826 

In the file picker control, the site file picker tab uses a `<DetailList>` for file selection. If a list of files is too long and requires pagination, it calls this `onRenderMissingItem` callback. This function is asyncronous and returns a promise, but this causes a React render error because `onRenderMissingItem` is expecting to render something.

Instead, we call the asyncronous function and return null.

### DIFF
--- a/src/controls/filePicker/controls/FileBrowser/FileBrowser.tsx
+++ b/src/controls/filePicker/controls/FileBrowser/FileBrowser.tsx
@@ -185,7 +185,7 @@ export class FileBrowser extends React.Component<IFileBrowserProps, IFileBrowser
                         selectionPreservedOnEmptyClick={true}
                         enterModalSelectionOnTouch={true}
                         onRenderRow={this._onRenderRow}
-                        onRenderMissingItem={this._loadNextDataRequest}
+                        onRenderMissingItem={() => { this._loadNextDataRequest(); return null; }}
                       />) :
                       (<TilesList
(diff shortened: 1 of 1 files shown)


AJIXuMuK:
thank you @cwparsons for the fix!
It has been merged and will be included in the next release!

PR #937, opened 2021-06-21
Feature - Improved Finnish translations

gautamdsheth:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [x]
| New sample?      | [ ]
| Related issues?  | NA

Improved Finnish translations. 
We checked with our team and found that these machine generated translations were somewhat misleading in Finnish so improved them

Added stock images missing keys for Finnish locale.
Looks like the stock images localization keys are missing for all languages except English, can you guys check that ?
Have added the correct one for Finnish.
Also, added Org Assets additional key for localization.


joelfmrodrigues:
@gautamdsheth many thanks! PS: I obviously don't know Finnish 😂

gautamdsheth:
Same here as well ! Our colleagues were kind enough to help me :) 
Btw any chance you can take a look at the stock images keys issue ? 
It is missing from non-English locales.

joelfmrodrigues:
@gautamdsheth sorry, had this on my inbox with a million other emails and only reading it now. Can you send/tag me on the issue you are referring to, if still open? 
I have had very limited time over the last few weeks, but can have a look when I get the chance

PR #944, opened 2021-06-26
Returning item instead of data for onSubmitted event

ravichandran-blog:
| Q               | A
| --------------- | ---
| Bugfix?        | [ X ]
| New feature?    | [ ]
| New sample?      | [ ]

Returning item instead of data for the submitted event and some corrections in the documentation

### DIFF
--- a/docs/documentation/docs/controls/DynamicForm.md
+++ b/docs/documentation/docs/controls/DynamicForm.md
@@ -8,7 +8,7 @@ This control can dynamically generate SharePoint list or SharePoint document lib
 - Import the following modules to your component:
 
 ```TypeScript
-import { DynamicForm } from "fx-controls-react/lib/DynamicForm";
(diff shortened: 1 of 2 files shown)


AJIXuMuK:
Hi @ravichandran-blog,

I have intentionally changed the returning value from PnPJS `item` instance to the data.
Some developers may not use PnPJS and want just the data for the created/modified item. We, as a library, should support this scenario.


Potentially we can return the PnPJS item as a second parameter or based on additional property.
What do you think @joelfmrodrigues @estruyf ?

joelfmrodrigues:
Sounds good to me 👍 Although everyone really should use PnPjs as it's great, it's their choice at the end of the day and I agree that we should not "enforce it".

AJIXuMuK:
I have added a new property `returnListItemInstanceOnSubmit`. By default it is `true`.
If it is set the second parameter of `onSubmitted` will contain an instance of PnPJS `IItem`

Issue #879, opened 2021-04-16
TaxonomyPicker suggestions language 


janijaak:
#### Category
[x] Enhancement

[ ] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [2.6.0]

#### Expected / Desired Behavior
Currently TaxonomyPicker **suggestions** language is always english/term store default language. It should always be the same as in the picker which seems to respect users selected language.

#### Observed Behavior
I'm using TaxonomyPicker in my form and when I start writing the term to the control I have to use English. When I select the term from picker the terms are shown in my users language which is Finnish.

In the picture you can also see that when the term was picked by writing in the TaxonomyPicker I had to use English. After that when I open the picker the term is in Finnish and also selected.

<img width="500" alt="TaxonomyPicker" src="https://user-images.githubusercontent.com/5503848/115009771-4b07be80-9eb5-11eb-9c24-851bf74abe93.png">

joelfmrodrigues:
@janijaak thanks for the suggestion. Would you be interested in implementing this change and submitting a PR?

petkir:
Hi @janijaak , Hi @joelfmrodrigues 

in PR #950  it is with an additional property solved ```useUserLanguage?:boolean```, but I think @janijaak  enhancement should be the default behavior. 

What do you think about changing to user UI language by default?

v1 and v2 are waiting for your response.

PR #950, opened 2021-07-05
879 dev

petkir:
| Q               | A
| --------------- | ---
| New feature?    | [ X]
| Related issues?  |  partially #879 

Termstore picker can now use the UI language of the user.

Details:
With the new Control Property useUserLanguage you can change the suggestion result language to the user Default Language.


petkir:
I will change the default behavior for the suggestions so I will remove the additional Property

Issue #958, opened 2021-07-13
[ PeoplePicker ] Default selected items for groups


a1exymoroz:
#### Category

[x] Bug

[x] Question

#### Version

Please specify what version of the library you are using: [ "@pnp/spfx-controls-react": "^3.2.1" ]

#### Question

Why doesn't PeoplePicker have selected items for different types?
PeoplePicker has only defaultSelectedUsers for users. But if we use different types ([PrincipalType.SecurityGroup, PrincipalType.User, PrincipalType.SharePointGroup]), I'll have problem.

#### Observed Behavior

If we can use different types, we have to select different types.

#### Steps to Reproduce

1) Set different  principalTypes
`  principalTypes={[PrincipalType.SecurityGroup, PrincipalType.User, PrincipalType.SharePointGroup]}`
2) Set different selected items
`defaultSelectedUsers={[User, Sharepoint Group, Azure AD group]}`
3) If we use `defaultSelectedUsers`, we need emails. But Sharepoint group doesn't have a email.
If I use texts for `defaultSelectedUsers`, Azure AD group won't select

Thanks!

AJIXuMuK:
Thanks @a1exymoroz for submitting both issue and fix!

PR #962, opened 2021-07-19
feat/#958 Add selected items for groups

a1exymoroz:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [ x ]
| New sample?      | [ ]
| Related issues?  | #958 

- Description was fixed
- Security group secondary text  returns email of group

### DIFF
--- a/src/services/PeopleSearchService.ts
+++ b/src/services/PeopleSearchService.ts
@@ -228,7 +228,7 @@ export default class SPPeopleSearchService {
             switch (element.EntityType) {
               case 'User':
                 const accountName: string =  element.Description || "";
-                const email: string  =  element.EntityData.Email || element.Description;
+                const email: string  =  element.EntityData?.Email || element.Description;
                 return {
                   id: element.Key,
                   loginName: element.LoginName ? element.LoginName : element.Key,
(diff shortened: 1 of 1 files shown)


gautamdsheth:
@a1exymoroz - This is a breaking change. So,  could you please modify the PR to make it so that `defaultSelectedUsers` and `defaultSelectedItems` are both supported ? 

If `defaultSelectedUsers` are specified, just append it to `defaultSelectedItems` array internally. 
Once we move to maybe v4, we can remove the `defaultSelectedUsers` property.

a1exymoroz:
> @a1exymoroz - This is a breaking change. So, could you please modify the PR to make it so that `defaultSelectedUsers` and `defaultSelectedItems` are both supported ?
> 
> If `defaultSelectedUsers` are specified, just append it to `defaultSelectedItems` array internally.
> Once we move to maybe v4, we can remove the `defaultSelectedUsers` property.

Ok. I'll change the PR

AJIXuMuK:
Sorry for not looking into this PR earlier.

I don't think we need to provide some new property. Why don't just use `defaultSelectedUsers` as `defaultSelectedItems`?
Yes, the name is not the best one and we can change it in the next major version. But 2 separate properties will bring more confusion.

a1exymoroz:
> Sorry for not looking into this PR earlier.
> 
> I don't think we need to provide some new property. Why don't just use `defaultSelectedUsers` as `defaultSelectedItems`?
> Yes, the name is not the best one and we can change it in the next major version. But 2 separate properties will bring more confusion.

I can revert to the first commit

Actually I need only one changes in 

image

Without this changes I can't set default selected items for security group

AJIXuMuK:
@a1exymoroz - yes, I agree that this change is needed (maybe with modification to check if `element.EntityData` is defined). But I think additional property brings confusion.

AJIXuMuK:
Thank you @a1exymoroz for the change. And sorry for back-and-forth communication!

PR #994, opened 2021-08-18
Fix - missing translation keys

gautamdsheth:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [x]
| New sample?      | [ ]
| Related issues?  | fixes NA

This PR adds the missing translation keys for non-English languages.

@joelfmrodrigues , @AJIXuMuK - this does sound like a big PR , looking at number of files. But it has only added missing translation keys, didn't delete any keys.

I just ran the translation script with my cognitive services key and it has added the missing keys, to view the changes maybe just use the ignore whitespaces option so that it only shows the changed values


joelfmrodrigues:
Many thanks @gautamdsheth 
I tried the view ignoring spaces and it definitely helps. Will try to review soon

joelfmrodrigues:
@gautamdsheth Many thanks again for this and apologies for the delay...
I noticed that some entries were still not translated (for example MyTeamsMessageDontHaveTeams) so feel free to send a new pull request if you can :)
Merging this one - Thanks!

PR #1003, opened 2021-08-30
Dev 879 TaxonomyPicker suggestions language

### LINKED ISSUE #879: TaxonomyPicker suggestions language

petkir:
| Q               | A
| --------------- | ---
| Bug fix?        | [ X]
| Related issues?  | fixes #879 (v3)

Taxonomy Selection Language Suggestions based on the user's language


AJIXuMuK:
Thank you @petkir for the PR.
It has been merged with small changes, and will be included in the next release.

PR #1006, opened 2021-08-31
fix(TaxonomyPicker): errorMessage not cleared when set to empty value

### LINKED ISSUE #953: TaxonomyPicker errorMessage label not being removed
#### Category
[ ] Enhancement

[x] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [latest]

When using the errorMessage prop and reverting it to an empty string, it should remove the previous errorMessage

#### Observed Behavior
When clearing the errorMessage prop from the component, the previous errorMessage is not cleared. The componentWillReceiveProps method escapes the assignation of the new errorMessage only when there is a value.

[TaxonomyPicker](https://github.com/pnp/sp-dev-fx-controls-react/blob/66f0362252e6bce8e5d4ded5eb430d72599abfc6/src/controls/taxonomyPicker/TaxonomyPicker.tsx#L93-L99)
```
public componentWillReceiveProps(nextProps: ITaxonomyPickerProps) {
    let newState: ITaxonomyPickerState | undefined;
    // Check if the initial values objects are not equal, if that is the case, data can be refreshed
    if (!isEqual(this.props.initialValues, nextProps.initialValues)) {
      newState = {
        activeNodes: nextProps.initialValues || []
      };
    }

    if (nextProps.errorMessage) {
      if (!newState) {
        newState = {};
      }

      newState.errorMessage = nextProps.errorMessage;
    }

    if (newState) {
      this.setState(newState);
    }
  }
```

#### Steps to Reproduce

1. Create a component with the TaxonomyPicker
2. Create a state with an errorMessage variable containing a non-empty random string
3. Assign the errorMessage prop of the TaxonomyPicker with the state value previously assigned
4. Create a button with an onClick that setState of the errorMessage with an empty string
5. Notice that the errorMessage from the TaxonomyPicker is not cleared

Fix suggestion : 
if (nextProps.errorMessage !== this.props.errorMessage) at line 93

Thanks!

umaknow-jeanluc:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #953

This PR fixes the TaxonomyPicker to remove the errorMessage label when the errorMessage props is cleared (empty string).
Also updated the ControlTest component to reflect that change.

image
image

### DIFF
--- a/src/controls/taxonomyPicker/TaxonomyPicker.tsx
+++ b/src/controls/taxonomyPicker/TaxonomyPicker.tsx
@@ -91,7 +91,7 @@ export class TaxonomyPicker extends React.Component<ITaxonomyPickerProps, ITaxon
       };
     }
 
-    if (nextProps.errorMessage) {
+    if (nextProps.errorMessage !== this.props.errorMessage) {
       if (!newState) {
         newState = {};
       }
--- a/src/webparts/controlsTest/components/ControlsTest.tsx
+++ b/src/webparts/controlsTest/components/ControlsTest.tsx
@@ -445,6 +445,7 @@ export default class ControlsTest extends React.Component<IControlsTestProps, IC
       showErrorDialog: false,
       selectedTeam: [],
       selectedTeamChannels: [],
+      errorMessage: "This field is required"
 
     };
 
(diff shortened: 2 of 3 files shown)


AJIXuMuK:
Thank you @umaknow-jeanluc for the PR.

PR #1017, opened 2021-09-22
Add multi numbers support for baseTemplate option in ListPicker

### LINKED ISSUE #1016: <ListPicker> baseTemplate option should allow string or multiple numbers
#### Category
[x] Enhancement

#### Version

Please specify what version of the library you are using: [ Latest : 3.3.0]

ListPicker Component should allow input of a string or multiple numbers because filter request is allowing it.
(Example : ?$filter=(BaseTemplate eq 100 or BaseTemplate eq 106))

#### Observed Behavior
ListPicker actually do not allow other input than number.

#### Steps to Reproduce
Use ListPicker component and set baseTemplate option.

Xiyitifu:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [x]
| New sample?      | [ ]
| Related issues?  | fixes #1016 

I've added ability to set an array to baseTemplate option of listpicker


joelfmrodrigues:
@Xiyitifu many thanks for the PR. Even though I personally prefer the way code it's formatted in the PR, as a general rule, we ask people to not change the formatting of code that they are not modifying. I suspect this could have been done automatically by a VS Code extension or even by accident, but could you please revert the formatting changes for code that was not modified?

AJIXuMuK:
I agree with @joelfmrodrigues - too many "refactoring" changes that make it difficult to review the actual new feature.

PR #1028, opened 2021-10-01
#1022 - fix issue with file name extension

### LINKED ISSUE #1022: fileNameWithoutExtension in FilePicker, not calculated right
#### Category
[ ] Enhancement
[X] Bug
[ ] Question
Please specify what version of the library you are using: [  3.3.0  ]

fileNameWithoutExtension for file name '01.06%20Malpractice%20Coverage_454342.docx'
should be '01.06%20Malpractice%20Coverage_454342' but the result shows '01'
It seems to substring until the first '.' which is wrong if there are more than one periods in the name.
It should be calculated using the LAST (rightmost) period in the name

Thanks!

gautamdsheth:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #1022 

Fix issue with file name containing `.` in the name. This assumes that the last dot will be the extension instead of the first dot implemented currently.

### DIFF
--- a/src/common/utilities/GeneralHelper.ts
+++ b/src/common/utilities/GeneralHelper.ts
@@ -266,9 +266,7 @@ export class GeneralHelper {
      */
     public static getFileNameWithoutExtension(itemUrl : string) {
       const fileNameWithExtension = GeneralHelper.getFileNameFromUrl(itemUrl);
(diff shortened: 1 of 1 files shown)


AJIXuMuK:
Thanks @gautamdsheth for the fix!

PR #1074, opened 2021-12-09
Parsing DynamicForm defaultValue for Boolean fields - Fix for #1073

### LINKED ISSUE #1073: Boolean fields in DynamicForm do not honour the default value in list settings
#### Category
[ ] Enhancement

[*] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [3.5.0]

#### Expected Behavior

When using the DynamicForm component, Boolean (Yes/No) fields should render the `default value` set in list settings. 

#### Observed Behavior

If a default value of "No" (false) is set in the list, the control still renders a default state of "Yes" (true).  

https://github.com/pnp/sp-dev-fx-controls-react/blob/master/src/controls/dynamicForm/DynamicForm.tsx

#### Steps to Reproduce

1. Add a list to SharePoint, and add Yes/No columns with default values of "Yes" and "No" respectively. 
2. Amend [ControlsTest.tsx#L886](https://github.com/pnp/sp-dev-fx-controls-react/blob/master/src/webparts/controlsTest/components/ControlsTest.tsx#L886) to point to this list, and _remove_ the `listItemId` attribute (causing the control to render a New Item form)
3. Observe that Boolean fields that should default to "No" actually default to "Yes"

#### Proposed solution

This appears to be caused by the `@fluentui` `Toggle` control receiving a `String` in its `defaultChecked` prop. Automatic type conversion of a `String` (of any length > 0) to `Boolean` will always be `true` - therefore, when this control is passed a `defaultValue` of "0" - this is treated as a truthy value.

Such values can be coerced to the correct boolean values using either `JSON.parse` or the `Boolean` and `Number` constructors. For example:

```js
const exampleA = JSON.parse("0"); // exampleA = 0
const exampleB = Boolean(Number("0")); // exampleB = false
```

Doing either of these would work. The fix could be placed in the private [getFieldInformations](https://github.com/pnp/sp-dev-fx-controls-react/blob/master/src/controls/dynamicForm/DynamicForm.tsx#L292) method of [DynamicForm.tsx](https://github.com/pnp/sp-dev-fx-controls-react/blob/master/src/controls/dynamicForm/DynamicForm.tsx). 

This method is already pre-processing default values that have been returned from the API - and there are multiple `if...else` statements for various field types, but not for the `Boolean` field type.

The simplest fix (1 line) is to amend [#L441](https://github.com/pnp/sp-dev-fx-controls-react/blob/master/src/controls/dynamicForm/DynamicForm.tsx#L441) to:

```ts
        else if (fieldType === "Location" || fieldType === "Boolean") {
          defaultValue = JSON.parse(defaultValue);
        }
```

However, if a more explicit and easy to understand fix would be preferred:

```ts
        else if (fieldType === "Location") {
          defaultValue = JSON.parse(defaultValue);
        }
        else if (fieldType === "Boolean") {
          defaultValue = Boolean(Number(defaultValue));
        }
```

I have tested both approaches and they work, and would be happy to submit a PR but I imagine this is easy to include if you're already working on this.

t0mgerman:
| Q               | A
| --------------- | ---
| Bug fix?        | [*]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #1073

Lines 444-446 of DynamicForm.tsx parse the string-based defaultValue, converting it to a Boolean. This fixes an issue whereby the DynamicForm control wasn't honouring the default value setting in list fields when presenting a 'New' form. See #1073

### DIFF
--- a/src/controls/dynamicForm/DynamicForm.tsx
+++ b/src/controls/dynamicForm/DynamicForm.tsx
@@ -441,6 +441,9 @@ export class DynamicForm extends React.Component<IDynamicFormProps, IDynamicForm
         else if (fieldType === "Location") {
           defaultValue = JSON.parse(defaultValue);
         }
+        else if (fieldType === "Boolean") {
+          defaultValue = Boolean(Number(defaultValue));
+        }
 
         tempFields.push({
           newValue: null,


AJIXuMuK:
Thank you @t0mgerman for the fix!

PR #1078, opened 2021-12-09
Amend DynamicForm.tsx to use EntityPropertyName rather than InternalName - Fix for #1077

### LINKED ISSUE #1077: DynamicForm does not load or save correctly when field name starts with special character
#### Category
[ ] Enhancement

[*] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [3.5.0]

#### Expected Behavior 

DynamicForm should display and save all field values, regardless of internal field name in SharePoint

#### Observed Behavior

When a field is created in SharePoint with a special character at the start of the field name, SharePoint replaces this character with a unicode representation like `\_x####\_` - where #### is the unicode character reference. For example: if you name a field "% Complete" - SharePoint will call this field "_x0025_Complete" internally. To complicate matters further, the property returned from the API will be named "OData__x0025_Complete". At present, because of this difference, fields named in this way will not work in DynamicForm.

This does not affect fields where a special character is used in the middle or at the end of the field name. Only those used at the beginning.

#### Steps to Reproduce

1. Add a SharePoint field to a list, beginning with a special character such as %
2. Create or edit a list item to ensure that this field has a value
3. Amend [ControlsTest.tsx#L886](https://github.com/pnp/sp-dev-fx-controls-react/blob/master/src/webparts/controlsTest/components/ControlsTest.tsx#L886) to uncomment and display the DynamicForm control, point this control at the list and list item, then view the test in Workbench
4. Observe that the list item value does not load correctly, and values cannot be saved back to the list

#### Proposed Solution

The necessary data to fix this issue is already being retrieved. Instead of using `field.InternalName`, use `field.EntityPropertyName`. In most cases this matches `field.InternalName` - but it _also matches_ the field property name as the SharePoint API expects it. 

So on [DynamicForm.tsx#L326](https://github.com/pnp/sp-dev-fx-controls-react/blob/master/src/controls/dynamicForm/DynamicForm.tsx#L326) use:

```ts
defaultValue = item[field.EntityPropertyName];
```

and on [DynamicForm.tsx#L458](https://github.com/pnp/sp-dev-fx-controls-react/blob/master/src/controls/dynamicForm/DynamicForm.tsx#L458) use:

```ts
columnInternalName: field.EntityPropertyName,
```

t0mgerman:
| Q               | A
| --------------- | ---
| Bug fix?        | [*]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #1077

Fixes an issue where the DynamicForm control would not load or save field values correctly if field(s) were created with special characters at the start of their display name i.e. "% Complete"

Amends DynamicForm.tsx to swap `field.InternalName` references for `field.EntityPropertyName` - which better reflects what the SharePoint API expects as property names. See #1077 for details.

### DIFF
--- a/src/controls/dynamicForm/DynamicForm.tsx
+++ b/src/controls/dynamicForm/DynamicForm.tsx
@@ -323,7 +323,7 @@ export class DynamicForm extends React.Component<IDynamicFormProps, IDynamicForm
         let dateFormat: DateFormat | undefined;
         let principalType = "";
         if (item !== null) {
-          defaultValue = item[field.InternalName];
+          defaultValue = item[field.EntityPropertyName];
         }
         else {
           defaultValue = field.DefaultValue;
@@ -455,7 +455,7 @@ export class DynamicForm extends React.Component<IDynamicFormProps, IDynamicForm
           context: this.props.context,
           disabled: this.props.disabled || (disabledFields && disabledFields.indexOf(field.InternalName) > -1),
           listId: this.props.listId,
-          columnInternalName: field.InternalName,
+          columnInternalName: field.EntityPropertyName,
           label: field.Title,
           onChanged: this.onChange,
           required: field.Required,


AJIXuMuK:
Thank you @t0mgerman for the fix!

Issue #1094, opened 2022-01-17
DateTimePicker switch allowTextInput


RicoNL:
#### Category
[X] Enhancement

[ ] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [  v3.5.0       ]

_If you are not using the latest release, please update and see if the issue is resolved before submitting an issue._

#### Expected / Desired Behavior / Question
Make allowTextinput a setting to switch on/off

This was previously requested in https://github.com/pnp/sp-dev-fx-controls-react/issues/928 but has been implemented as always on

![image](https://user-images.githubusercontent.com/35333175/149778096-4a77d814-362d-46fc-b3f8-759cd6bd6658.png)

#### Observed Behavior
Not possible to switch off manual input of date.

#### Steps to Reproduce
Use the date picker control

RicoNL:
Thanks!

PR #971, opened 2021-07-23
Cherry pick of PR #951

modern-dev-dude:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [x ]
| New sample?      | [ ]
| Related issues?  | Adds enhancement  #928 , partially #Y, mentioned in #Z

#### What's in this Pull Request?

set allowTextInput to true
added handleTextChange for specific error message when entering a string
added loc string for english textErrorMessage 
added DateTimePickerTextErrorMessage to the IDateTimePickerInterface
added missing dependency for unit testing used in setup.js
added init test for date time picker to mount, render, update, and unmount

***

PR #1107, opened 2022-01-29
Added allowTextInput property

milanholemans:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [ x ]
| New sample?      | [ ]
| Related issues?  | fixes [#1094](https://github.com/pnp/sp-dev-fx-controls-react/issues/1094)

Exposed the allowTextInput property. Developer can now choose whether the user can add dates as text value rather than picking one from the date picker.

### DIFF
--- a/docs/documentation/docs/controls/DateTimePicker.md
+++ b/docs/documentation/docs/controls/DateTimePicker.md
@@ -62,6 +62,7 @@ The `DateTimePicker` control can be configured with the following properties:
 | isMonthPickerVisible | boolean | no | Controls whether the month picker is shown beside the day picker or hidden. |
 | showMonthPickerAsOverlay | boolean | no | Show month picker on top of date picker when visible. |
 | showWeekNumbers | boolean | no | Controls whether the calendar should show the week number (weeks 1 to 53) before each week row |
+| allowTextInput | boolean | no | Whether the user is allowed to enter a date instead of picking one from the date picker. |
 | strings | IDatePickerStrings | no | Localized strings to use in the DateTimePicker |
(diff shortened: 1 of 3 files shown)


joelfmrodrigues:
@milanholemans many thanks for the PR. I just noticed that the previous default value was `true` and now is `false` so wondering if this could cause issues to people who upgrade and don't notice the change. Could this be set to `true` by default, or did you experience any issues?

milanholemans:
> @milanholemans many thanks for the PR. I just noticed that the previous default value was `true` and now is `false` so wondering if this could cause issues to people who upgrade and don't notice the change. Could this be set to `true` by default, or did you experience any issues?

The thing is, I noticed that the DatePicker component of office-fabric-ui uses `false` as default value. That's why I made it false here by default as well. I can change it to `true` if needed.

joelfmrodrigues:
@AJIXuMuK what are your thoughts regarding changing this default value? Should we avoid it or go ahead and ensure that it's documented/announced?

milanholemans:
It has been changed from `false` to `true` last year, check this PR [#971](https://github.com/pnp/sp-dev-fx-controls-react/pull/971).

AJIXuMuK:
@joelfmrodrigues I would not change the defaults without a necessity. 
@milanholemans - thank you for providing this info. 
This is our mistake. We should have implemented this as property right away. 
But for now - let's leave the default as is.

milanholemans:
@AJIXuMuK @joelfmrodrigues I changed the default value to `true`.

joelfmrodrigues:
@milanholemans many thanks, it has now been merged

Issue #1096, opened 2022-01-17
NEW CONTROL - Adaptive Card Host


fabiofranzini:
Hello,
I've been working with Adaptive Cards, I really like the new ACE model and in general I like the fact that I can describe the UI of a Web Part in a declarative way and be able to change it at runtime, like a configuration.

After some tests on this, **I created a React control that allows you to render an Adaptive Card as a component**, also using the templating part and the ability to invoke commands defined using the "Action.Execute" model.

I would like to make it a control of this library, what do you think?

joelfmrodrigues:
Hi @fabiofranzini , that would be a great addition in my opinion. I was actually discussing this with @joaojmendes a few days ago as he was considering a similar control. Perhaps we could use this thread as a starting point to avoid duplication of effort and duplicated PRs?

Are you both able to describe your approach so we could all brainstorm the best way to move this forward and hopefully benefit from both solutions?

fabiofranzini:
Hi @joelfmrodrigues,
in recent days I chatted with @joaojmendes about this thing, we can join forces and bring the best of our solutions.

The idea is to have a control that can host Adaptive Cards in both Web Parts and Application Customizer, so that they can be used as a UI.

Also, as I have already implemented in some examples, we could extend the Adaptive Cards 'standard' set of controls with Fluent UI controls or as suggested by @joaojmendes with Microsoft Graph Toolkit controls.

@joaojmendes , we need to set up a call to discuss this :)

fabiofranzini:
Hi @AJIXuMuK, yes and thanks for the suggestion! The purpose is to create a reusable control and extend the common set of controls that come from Adaptive Cards.

joaojmendes:
Hi, all,

I think that we can accept the control of @fabiofranzini and the new improvements will be implemented on that .

Thanks
João

From: Fabio Franzini ***@***.***>
Date: Tuesday, 18 January 2022 at 16:24
To: pnp/sp-dev-fx-controls-react ***@***.***>
Cc: João Mendes ***@***.***>, Mention ***@***.***>
Subject: Re: [pnp/sp-dev-fx-controls-react] NEW CONTROL - Adaptive Card Host (Issue #1096)

Hi @AJIXuMuK<https://github.com/AJIXuMuK>, yes and thanks for the suggestion! The purpose is to create a reusable control and extend the common set of controls that come from Adaptive Cards.

fabiofranzini:
[...]

PR #1119, opened 2022-02-10
Added 'AdaptiveCardHost' Control

fabiofranzini:
Added 'AdaptiveCardHost' Control and Documentation

| Q               | A
| --------------- | ---
| Bug fix?        | [no]
| New feature?    | [yes]
| New sample?      | [no]
| Related issues?  | mentioned in #1096 

I added the new 'AdaptiveCardHost' Control used for rendering Adaptive Cards in React using integrated features likes SP and Teams Themes, using Fluent UI React Controls for Elements and Actions and Templating.


src/controls/adaptiveCardHost/fluentUI/Actions.tsx
```diff
@@ -0,0 +1,231 @@
+import { ActionButtonState, ActionIconPlacement, defaultHostConfig, ExecuteAction, OpenUrlAction, ShowCardAction, SubmitAction, ToggleVisibilityAction } from "adaptivecards";
+import { BaseButton, Button, CompoundButton, createTheme, DefaultButton, ITheme, PrimaryButton } from "office-ui-fabric-react";
+import * as React from "react";
+import * as ReactDOM from "react-dom";
+import { getDefaultFluentUITheme, hostCapabilitiesFluentUIThemeKey } from "./FluentUI";
+import { createDiv } from "./Shared";
+const redFluentUIPalette = require("../themes/redPalette.json");
+
+export class FluentUIExecuteAction extends ExecuteAction {
```
joelfmrodrigues, line 9: The first 5 classes (FluentUIExecuteAction, FluentUIOpenUrlAction, FluentUIShowCardAction, FluentUISubmitAction, FluentUIToggleVisibilityAction) seem to have the exact same code for `actionClickHandler` and `render` methods. 
If they will likely remain the same in the future, could this be refactored to avoid duplicated code?

src/controls/adaptiveCardHost/fluentUI/Actions.tsx
```diff
@@ -0,0 +1,231 @@
+import { ActionButtonState, ActionIconPlacement, defaultHostConfig, ExecuteAction, OpenUrlAction, ShowCardAction, SubmitAction, ToggleVisibilityAction } from "adaptivecards";
+import { BaseButton, Button, CompoundButton, createTheme, DefaultButton, ITheme, PrimaryButton } from "office-ui-fabric-react";
```
joelfmrodrigues, line 2: Can you please update imports from office-ui-fabric-react to be specific, so it aligns with the other library controls?
for example: 'office-ui-fabric-react/lib/Button'

Same for imports in the other files

src/controls/adaptiveCardHost/fluentUI/Elements.tsx
```diff
@@ -0,0 +1,838 @@
...
+        }
+        else {
+            return undefined;
+        }
+    }
+}
+
+export class FluentUITimeInput extends Input {
+    private static convertTimeStringToDate(timeString: string): Date {
+        return new Date("1973-09-04T" + timeString + ":00Z");
```
joelfmrodrigues, line 621: Is this a hardcoded date from testing?

joelfmrodrigues:
Trying to type on the Name field of the test control seems to behave strangely for the first characters typed as they sometimes disappear slightly after. The control also seems to lose focus the first time. Could this be a problem or likely caused by something else on the page (there's a lot going on there with the other controls)? 
image

joelfmrodrigues:
Clicking the 'Show Card' button on the test control gives an error, but works if clicked again. The same for closing the card using the button: clicking the first time gives an error, the second time works.
image

joelfmrodrigues:
> Trying to type on the Name field of the test control seems to behave strangely for the first characters typed as they sometimes disappear slightly after. The control also seems to lose focus the first time. Could this be a problem or likely caused by something else on the page (there's a lot going on there with the other controls)? image

I think we can ignore this one as I believe this was caused by the Carousel control. After removing the 3 instances of the Carousel control from the page this issue didn't seem to happen again.

joelfmrodrigues:
@fabiofranzini I added some comments for some minor things 🙂

fabiofranzini:
Thanks @joelfmrodrigues I'll check this ASAP 🙂

src/controls/adaptiveCardHost/fluentUI/Actions.tsx
fabiofranzini: Ok, I'll check in all files and I'll change that

src/controls/adaptiveCardHost/fluentUI/Actions.tsx
fabiofranzini: Ok, I'll check in all files and I'll change that

fabiofranzini: Done.

src/controls/adaptiveCardHost/fluentUI/Elements.tsx
fabiofranzini: Yes, it's for comparing time only.
It's the same code present into the default implementation: https://github.com/microsoft/AdaptiveCards/blob/f764013a76899b6776291edd1de1cbe1506e7faf/source/nodejs/adaptivecards/src/card-elements.ts#L4032

src/controls/adaptiveCardHost/fluentUI/Actions.tsx
fabiofranzini: Yes, I know! Do you have any suggestions on how to resolve this?

fabiofranzini: The problem is this: Every Actions (ExecuteAction, OpenUrlAction, ShowCardAction, SubmitAction, ToggleVisibilityAction) have a base class named Action that is an abstract class... I'm not able to create a generic class that extends an abstract class... this is the reason I decided to create 5 classes whit the same code... If you have a suggestion, it will very appreciated :)

joelfmrodrigues: I was just looking into it not and found the same issues 😅 didn't realise that when I added the comment. I will have another quick look but this can go as is if we can't find an elegant solution

joelfmrodrigues: Nope, can't think of a better solution so leave it as is. Sorry for that

src/controls/adaptiveCardHost/fluentUI/Actions.tsx
joelfmrodrigues: @fabiofranzini can you double-check that you have committed and pushed changes?

AJIXuMuK (changes requested):
@fabiofranzini - please check out some additional comments I left.
Thanks for the great work!

fabiofranzini:
> Clicking the 'Show Card' button on the test control gives an error, but works if clicked again. The same for closing the card using the button: clicking the first time gives an error, the second time works. image

I'm not able to reproduce this bug (whether I use the test webpart with all controls or leaving only the AdaptiveCardHost control.) ... very strange...

ShowCardDemostration

I'm trying to remove the useless code from 'updateCssClasses' function and implement like this:
```
protected updateCssClasses() {
        if (this.renderedElement) {
            this.renderedElement.setAttribute(
                "aria-expanded",
                (this.state === ActionButtonState.Expanded).toString()
            );
        }
    }
```

src/controls/adaptiveCardHost/AdaptiveCardHost.css
```diff
@@ -0,0 +1,142 @@
...
+    font-size: 14px;
+    font-weight: 600;
+    padding: 4px 10px 5px 10px;
+    -webkit-user-select: none;
+    -moz-user-select: none;
+    -ms-user-select: none;
+    user-select: none;
+    height: 32px;
+    border-radius: 2px;
+    border: 1px solid var(--ach-fui-buttonBorder, #6290FF);
```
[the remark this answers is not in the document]
fabiofranzini: The variables are defined by calling a function in this lines:

1. https://github.com/fabiofranzini/sp-dev-fx-controls-react/blob/962e795894673745798d569d1114eeb47db9a349/src/controls/adaptiveCardHost/AdaptiveCardHost.tsx#L144
2. https://github.com/fabiofranzini/sp-dev-fx-controls-react/blob/962e795894673745798d569d1114eeb47db9a349/src/controls/adaptiveCardHost/AdaptiveCardHost.tsx#L151

I found into the W3C specification the possibility to use the fallback value: https://www.w3.org/TR/css-variables/
It's not a problem, the CSS vars are setted every time.

What I have to do in this case?

src/controls/adaptiveCardHost/AdaptiveCardHost.tsx
```diff
@@ -0,0 +1,232 @@
...
+    // create executeAction
+    const invokeAction = useCallback((action: Action) => {
+        if (props.onInvokeAction) {
+            let actionResult: IAdaptiveCardHostActionResult;
+            const type = action.getJsonTypeName();
+            switch (type) {
+                case OpenUrlAction.JsonTypeName: {
+                    let typedAction = action as OpenUrlAction;
+                    actionResult = {
+                        type: typedAction.getJsonTypeName(),
```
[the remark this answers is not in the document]
fabiofranzini: Done

joelfmrodrigues:
> > Clicking the 'Show Card' button on the test control gives an error, but works if clicked again. The same for closing the card using the button: clicking the first time gives an error, the second time works. image
> 
> I'm not able to reproduce this bug (whether I use the test webpart with all controls or leaving only the AdaptiveCardHost control.) ... very strange...
> 
> ShowCardDemostration
> 
> I'm trying to remove the useless code from 'updateCssClasses' function and implement like this:
> 
> ```
> protected updateCssClasses() {
>         if (this.renderedElement) {
>             this.renderedElement.setAttribute(
>                 "aria-expanded",
>                 (this.state === ActionButtonState.Expanded).toString()
>             );
>         }
>     }
> ```

@fabiofranzini think I have found the glitch with the error on button click 🙂 seems to happen when running using `npm run serve`, but not with `gulp serve`. Are you using gulp serve?

fabiofranzini:
> > > Clicking the 'Show Card' button on the test control gives an error, but works if clicked again. The same for closing the card using the button: clicking the first time gives an error, the second time works. image
> > 
> > 
> > I'm not able to reproduce this bug (whether I use the test webpart with all controls or leaving only the AdaptiveCardHost control.) ... very strange...
> > ShowCardDemostration
> > I'm trying to remove the useless code from 'updateCssClasses' function and implement like this:
> > ```
> > protected updateCssClasses() {
> >         if (this.renderedElement) {
> >             this.renderedElement.setAttribute(
> >                 "aria-expanded",
> >                 (this.state === ActionButtonState.Expanded).toString()
> >             );
> >         }
> >     }
> > ```
> 
> @fabiofranzini think I have found the glitch with the error on button click 🙂 seems to happen when running using `npm run serve`, but not with `gulp serve`. Are you using gulp serve?

@joelfmrodrigues  I'm using gulp serve... It's ok?

src/controls/adaptiveCardHost/AdaptiveCardHost.tsx
```diff
@@ -0,0 +1,232 @@
...
+                theme = createDarkTeamsTheme();
+            }
+                break;
+            case AdaptiveCardHostThemeType.TeamsHighContrast: {
+                theme = createHighContrastTeamsTheme();
+            }
+                break;
+        }
+
+        fluentUIThemeInstanceRef.current = convertFromPartialThemeToTheme(theme);
```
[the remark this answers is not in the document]
fabiofranzini: Nope, the problems using vars is described here: https://www.emgoto.com/storing-values-with-useref/#whats-the-difference-between-useref-and-using-a-variable

src/controls/adaptiveCardHost/AdaptiveCardHost.tsx
```diff
@@ -0,0 +1,232 @@
...
+    return (
+        <ThemeProvider theme={(fluentUIThemeInstanceRef.current) ? fluentUIThemeInstanceRef.current : getDefaultFluentUITheme()}>
+            <div
+                ref={renderElementRef}
+                className={`${(props.className) ? props.className : ""}`}
+                style={props.style}
+            />
+        </ThemeProvider>
+    );
+};
```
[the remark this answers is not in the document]
fabiofranzini: Yes, useEffects are written the way they will fire sequentially and with all the dependencies they need, that's why useRefs are used.

src/controls/adaptiveCardHost/AdaptiveCardHost.tsx
```diff
@@ -0,0 +1,232 @@
...
+} from 'react';
+import './AdaptiveCardHost.css';
+import { convertFromPartialThemeToTheme, createDarkTeamsHostConfig, createDefaultTeamsHostConfig, createHighContrastTeamsHostConfig, createSharePointHostConfig, initProcessMarkdown, setFluentUIThemeAsCSSVariables } from './AdaptiveCardHostHelpers';
+import { createDarkTeamsTheme, createDefaultTeamsTheme, createHighContrastTeamsTheme, getDefaultFluentUITheme, setFluentUIThemeAsHostCapability, useLocalFluentUI } from './fluentUI';
+import { IAdaptiveCardHostActionResult, IAdaptiveCardHostProps, AdaptiveCardHostThemeType } from './IAdaptiveCardHostProps';
+
+// Init Process Markdown
+initProcessMarkdown();
+
+export const AdaptiveCardHost = (props: IAdaptiveCardHostProps) => {
```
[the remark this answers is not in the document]
fabiofranzini: @AJIXuMuK Is destructuring mandatory?
In general, I prefer to understand which variables come from the props and which do not, by deconstructing this advantage is lost. 
If it is mandatory to deconstruct it, I do it, otherwise I prefer to leave it like this.
What do I have to do? 🙂

src/controls/adaptiveCardHost/AdaptiveCardHost.css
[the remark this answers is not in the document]
fabiofranzini: I removed the fallback from the code... fixed the way to apply the CSS vars ad the body element (to fix the problem with the list actions menu):

```JS
// just to fix the colors of the list actions menu
setFluentUIThemeAsCSSVariables(
    document.body,
    (props.isUniqueControlInPage) ? theme : getDefaultFluentUITheme()
);
// *****
```

joelfmrodrigues:
> > > > Clicking the 'Show Card' button on the test control gives an error, but works if clicked again. The same for closing the card using the button: clicking the first time gives an error, the second time works. image
> > > 
> > > 
> > > I'm not able to reproduce this bug (whether I use the test webpart with all controls or leaving only the AdaptiveCardHost control.) ... very strange...
> > > ShowCardDemostration
> > > I'm trying to remove the useless code from 'updateCssClasses' function and implement like this:
> > > ```
> > > protected updateCssClasses() {
> > >         if (this.renderedElement) {
> > >             this.renderedElement.setAttribute(
> > >                 "aria-expanded",
> > >                 (this.state === ActionButtonState.Expanded).toString()
> > >             );
> > >         }
> > >     }
> > > ```
> > 
> > 
> > @fabiofranzini think I have found the glitch with the error on button click 🙂 seems to happen when running using `npm run serve`, but not with `gulp serve`. Are you using gulp serve?
> 
> @joelfmrodrigues I'm using gulp serve... It's ok?

Sure, maybe a bug with spfx-fast-serve then. If it works with gulp serve then it's all good I think. As a side note, if you don't generally use fast serve I strongly advise you to give it a try as it's insanely fast

fabiofranzini:
@joelfmrodrigues and @AJIXuMuK I just committed the changes... Please check now and of course many thanks for your feedback 🙂

src/controls/adaptiveCardHost/AdaptiveCardHost.tsx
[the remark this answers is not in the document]
AJIXuMuK: `ref`s are important to store the value between calls and to avoid re-call of `useEffect`.
But in your case here you do `fluentUIThemeInstanceRef.current = convertFromPartialThemeToTheme(theme)` and then 4 lines lower you are using `fluentUIThemeInstanceRef.current` which will not change in between.
So why not do that
```typescript
const themeInstance = converFromPartialThemetoTheme(theme);
fluentUIThemeInstanceRef.current = themeInstance;
// case block
hostConfig = <call_you_method>(themeInstance);
```

You can leave as is but variable should be more performant.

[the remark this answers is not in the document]
fabiofranzini: @AJIXuMuK sorry I didn't understand your suggestion... ok, you are right, let me fix today 🙂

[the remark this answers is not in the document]
fabiofranzini: Done 🙂

fabiofranzini:
@joelfmrodrigues @AJIXuMuK let me know if I must Squash this three commits or not.
I added a small fix, in the last commit, the return value for 'get value()' prop for the FluentUIDateInput to the correct format. 🙂

AJIXuMuK (approved):

AJIXuMuK:
Thanks again @fabiofranzini for the great contribution!

fabiofranzini:
> Thanks again @fabiofranzini for the great contribution!

It's a pleasure! I'll do more contributions in the next months 🙂

Issue #1145, opened 2022-03-11
Add SPFx theme & context property to the AdaptiveCardHost control


fabiofranzini:
#### Category
[X] Enhancement
[-] Bug
[-] Question

#### Version
Please specify what version of the library you are using: [3.6.0]

#### Expected
Speaking with @joaojmendes about the possible evolutions of the control, and specifically in the creation of a library of additional elements (using MGT for example), it is necessary to pass the "Theme" object and values that come from the "Context" of SPFx, so to be used with the Adaptive Card "templating" syntax and therefore to be passed to the new elements.

The idea is to directly inject these values into the "Data" object which is then used to do the "binding" in the rendering phase.

An example could be to pass the colors of the theme applied to a template to discriminate which image to load (understand if the theme is dark or not), etc., etc.

Or, thanks to the context of SPFx, you can have information on the site, current user, etc., etc. and use that information as properties in the "data" object

joelfmrodrigues:
@fabiofranzini are you aiming to create a PR for this? If so, I can assign this item to you

fabiofranzini:
@joelfmrodrigues yes! I'll create the PR in 30 mins 😀

fabiofranzini:
@joelfmrodrigues and @joaojmendes done 😀

PR #1146, opened 2022-03-11
AdaptiveCardHost - Added "context" property

fabiofranzini:
| Q               | A
| --------------- | ---
| New feature?    | [X]
| Related issues?  | #1145 #1154

Added SPFx "context" property to "AdaptiveCardHost" control to inject "@context" field in the data object for access context and theme information inside the Adaptive Card using the templating model.


src/controls/adaptiveCardHost/AdaptiveCardHostHelpers.ts
```diff
@@ -1024,4 +1025,27 @@ const setCSSVariables = (domElement: HTMLElement, prefix: string, obj: any) => {
             domElement.style.setProperty(`--${prefix}-${key}`, obj[key]);
         });
     }
+};
+
+export const injectContextProperty = (dataObject: { "$root": object }, theme: ITheme, spfxContext: BaseComponentContext): IEvaluationContext => {
+    let evaluationContext: IEvaluationContext;
+    let context = {
+        theme: theme,
+        aadInfo: (spfxContext?.pageContext?.aadInfo) ? JSON.parse(JSON.stringify(spfxContext?.pageContext?.aadInfo)) : undefined,
```
AJIXuMuK, line 1034: `JSON.stringify` on classes may lead to issues in future if there some new properties-instances with circular dependencies.
Not ideal way, but not buggy, is to directly specify what properties from the class you want to copy to your context object.

AJIXuMuK (changes requested):

src/controls/adaptiveCardHost/AdaptiveCardHostHelpers.ts
fabiofranzini: Hi @AJIXuMuK , I understand your request of changes and you are right. Let me understand a different way to do that, and I'll push the updates.

fabiofranzini: Hi @AJIXuMuK , I followed your suggestion and fixed how to retrieve and use context properties. This way there are no problems with serialization. Thanks!

fabiofranzini:
Hi @AJIXuMuK , I removed the "spPartialLegacyContext" from my previous commit because it's doesn't have sense at this moment. The legacy context from SPFx PageContext is legacy and at this moment it's not thought to use it in any scenario.

I also added a new commit with the code to add the "telemetry" for this control. 🙂

fabiofranzini:
Hi @AJIXuMuK, can you check the last commit and give me some feedback on this code? I need to approve this PR before creating a new one 🙂

fabiofranzini:
Hi everyone, I'm implementing Adaptive Cards designer as a react control. to do this I'm thinking about how to modify and reuse this control. This PR was to introduce "direct" support to SharePoint context properties, directly in the "data" object.
This, however, after much reasoning, introduces a "link" with the context of SPFx, and if one does not want to use it in this way one would be forced to have to pass the SharePoint context anyway.

That said, I was struggling to change the PR again and introduce not a control property, but a helper that allows you to inject the same properties in a way like this:

```
let dataObject = {"a":1};
<AdaptiveCardHost
  ...
  data={injectSPFxContextData(dataObject)}
  ...
/>
```
before approving this PR, can you tell me your opinion? 🙂
/cc @AJIXuMuK @joelfmrodrigues @joaojmendes

AJIXuMuK:
@fabiofranzini - sorry for the long delay.
Maybe we could just make the `context` property optional?
You already have the logic to combine `data` and `context`.
If the `context` is optional, developers won't need to call any additional functions.

What do you think?

fabiofranzini:
> @fabiofranzini - sorry for the long delay. Maybe we could just make the `context` property optional? You already have the logic to combine `data` and `context`. If the `context` is optional, developers won't need to call any additional functions.
> 
> What do you think?

Hi @AJIXuMuK, no problems 🙂
Yes, this could be a good compromise...
Let me do this change today and I'll notify you for the review 🙂

fabiofranzini:
@AJIXuMuK done! 
I merged another fix from #1154 
Sorry for this, but I had already gone ahead with the change and then I found myself with the code already written.
For the next few times I will avoid doing this and will keep the code separate for each issue.
I have marked the "isUniqueControlInPage" property as obsolete and we no longer use it in the code, so those who have already used the control will not have problems compiling. 🙂

AJIXuMuK:
@fabiofranzini - `context` is still marked as required in the docs.
Also - could you please explain the usage of `nano-css`?
Why is that needed instead of component's `scss` or something from Fluent UI?

fabiofranzini:
@AJIXuMuK fixed the docs, thanks 🙂
I used `nano-css` to generate dynamically the class names for some styles to apply to the Adaptive Card HostConfig.
This is because the Adaptive Card HostConfig has a property called `cssClassNamePrefix`, and to set the rendering style (I need it mostly to solve the style problem in the Sub Actions menu) I need to have "dynamic" class names. Otherwise, if they were static, I would have the conflict problem if multiple instances of this control are on the same page (this is the way suggested by the Adaptive Cards product group to remove the `isUniqueControlInPage` property)

AJIXuMuK:
merged!

fabiofranzini:
> merged!

Thanks @AJIXuMuK!
Next step designer 😀

PR #1147, opened 2022-03-11
minor documentation fix

MathijsVerbeeck:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | 

Fixed a minor documentation error


joelfmrodrigues:
Hi @Mathizor , many thanks for the PR.
I can confirm that there is an issue with the documentation, but when checking the interface I can see that the property accepts both `string` and `JSX.Element`, so I think we need to escape the | character and have both properties
image

Can you please update the PR?

AJIXuMuK:
Thank you @MathijsVerbeeck for the fix!

PR #1160, opened 2022-03-21
Fix: sorting the terms in locale language

gautamdsheth:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]

This PR fixes an issue with terms not getting sorted when used in language specific browser with labels.
It always follows the default order of term store.

To repro this bug:

Create a term set with some terms. 
In the term store, ensure that you have atleast one other language available for use.
Add the label values in the other language.

It will always fetch the order of the default language


AJIXuMuK:

Issue #1125, opened 2022-02-12
LocationPicker width


PathToSharePoint:
#### Category
[ ] Enhancement

[ ] Bug

[x] Question

#### Version

Please specify what version of the library you are using: [3.5.0]

#### Expected / Desired Behavior / Question
How can I force the width of the LocationPicker? Or have it fit in its container?
I thought the components were built on Office UI Fabric, but I don't see a style prop.

#### Observed Behavior

Here is what it looks like in my test, I am unable to change the address once it is set:
![image](https://user-images.githubusercontent.com/40323257/153700282-a2aecc36-5cf3-4525-b82e-e451e8c7a75b.png)

Thanks!

joelfmrodrigues:
Hi @PathToSharePoint , I just tested this and there is no option to style.
Would you be interested in submitting a PR? It should be a simple change and similar to the properties available in the People picker control

joelfmrodrigues:
@PathToSharePoint that's great ☺️
I have assigned this issue to you (no pressure), but let me know if you change your mind and I can remove it. It's only to try to prevent more than one person working on it

PathToSharePoint:
I took a closer look, and here is the issue I am trying to address (no pun intended): if the address is too long, it pushes the delete control (cross) outside of the viewport, and it becomes impossible to edit the field.

I think the easiest way to fix the issue is to crop the address and append an ellipsis, like it's done in the list picker. Another way would be to display the address on multiple lines, like in the screenshot below.

It seems that the control is built on the Fluent UI Persona component, let me see what styling it allows.

![image](https://user-images.githubusercontent.com/40323257/159146382-4a67cc14-d88c-4ebf-a38e-297e9ec2f886.png)

PathToSharePoint:
Looking at the code, I found in `LocationPicker.module.scss` a class name `persona`. That should allow me to inject some additional styling, without major refactoring.

PathToSharePoint:
@ravichandran-blog  and @AJIXuMuK, I see that you are the authors of this control.

I plan to remove `width: fit-content;` in `LocationPicker.module.scss` as it seems to be what's causing the issue. Do you see any objection?
```
.pickerItemContainer {
  display: flex;
  border-style: solid;
  border-color: "[theme: themePrimary, default: #0078d7]";
  border-radius: 22px;
  min-width: 16px;
  border-width: 2px;
  width: fit-content;
}
``` 
[...]

PR #1165, opened 2022-03-22
Trigger onChange on picker clear action

PathToSharePoint:
In the current release, the onChange event is not triggered when the selected item is reset to null (Clear button).

| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | mentioned in #1125

Trigger the onChange event handler when the clear button is clicked and selected item reset to null.


AJIXuMuK:
Thank you @PathToSharePoint for the fix!

PR #1175, opened 2022-03-31
people picker transitive members

### LINKED ISSUE #1173: Peoplepicker: search users in nested security groups
#### Category
[X] Enhancement

[ ] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [ 3.7.0 ]

Make users from nested Security groups searchable when the groupId is specified in the PeoplePicker.

Changing 'members' to 'transitiveMembers' in PeopleSearchService.ts | searchTenant should do the trick. Ideally implement a switch to choose between members and transitiveMembers

image

#### Observed Behavior
 Currently only users directly added to the group are returned when searched for. Users added to a nested group are not found.

#### Steps to Reproduce
- Create a webpart with peoplepicker control
- Specify a groupid-string from a security group with both users fromnested security group and users directly added to the group. Only users directly added to the AD group are returned when searched for.

RicoNL:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [X]
| New sample?      | [ ]
| Related issues?  | fixes #1173

When groupId-string is provided now fetches also transitive members in case an id of a security group is provided. Previously only direct members of the security group were fetched.

### DIFF
--- a/src/services/PeopleSearchService.ts
+++ b/src/services/PeopleSearchService.ts
@@ -195,9 +195,9 @@ export default class SPPeopleSearchService {
         searchBody.queryParams["SharePointGroupID"] = groupId;
       }
 
-      // Check if users need to be searched in a specific Office365 Group
+      // Check if users need to be searched in a specific Microsoft 365 Group, Security Group (incl. nested groups) or Distribution List
       else if(groupId && typeof(groupId) === 'string') {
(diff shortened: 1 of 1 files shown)


AJIXuMuK:
Thank you @RicoNL for the enhancement!

PR #1232, opened 2022-06-13
feat(PeoplePicker): add new `allowUnvalidated` option to allow adding non-tenant users

JakeStanger:
Guest users can be identified by the `userUnvalidated` flag.

This is useful for then handling inviting users as guests.

| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [x]
| New sample?      | [ ]
| Related issues?  | 

Adds new prop to the `PeoplePicker` component called `allowUnvalidated`. Setting this to true no longer filters out users which are "unvalidated", which in general means they do not exist in the tenant. This allows the picker to be used for example for an interface to invite guest users.

If a user is unvalidated, the `isUnvalidated` key will be true on the returned object.

### DIFF
--- a/src/controls/peoplepicker/IPeoplePicker.ts
+++ b/src/controls/peoplepicker/IPeoplePicker.ts
@@ -114,6 +114,10 @@ export interface IPeoplePickerProps {
    */
   ensureUser?: boolean;
   /**
+   * When true, allow email addresses that have not been validated to be entered, effectively allowing any use
+   */
+  allowUnvalidated?: boolean;
+  /**
    * Placeholder to be displayed in an empty term picker
    */
   placeholder?: string;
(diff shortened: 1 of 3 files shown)


JakeStanger (commented):
Good spot, thanks :)

joelfmrodrigues:
Hi @JakeStanger , many thanks for the contribution. The PR has now been merged and changes will be available as part of the next release. You can also test the changes as part of the beta release that will be published automatically in a few minutes.

joelfmrodrigues:
Also, as a side note, I have updated the control docs to include the new property 🙂
Thanks again

JakeStanger:
Ah I completely forgot about docs, thanks!

PR #1242, opened 2022-06-22
Fix changelog script by setting CHANGELOG.JSON filename extension to lower case

danwatford:
| Q               | A
| --------------- | ---
| Bug fix?        | [X]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  |

The git pre-commit hook currently fails when running on hosts with case-sensitive filesystems.
Renaming the `CHANGELOG.JSON` file to use a lower case `.json` extension resolves this issue.


danwatford:
Hi @joelfmrodrigues , this should fix the issue you are seeing with the pre-commit hook.

joelfmrodrigues:
@danwatford good point, case sensitive 🙂
image
Thank you!

Issue #1241, opened 2022-06-21
LivePersona is not working in spfx project


muppanasashikumar:
Thank you for reporting an issue, suggesting an enhancement, or asking a question. We appreciate your feedback - to help the team understand your
needs please complete the below template to ensure we have the details to help. Thanks!

**Please check out the [documentation](https://pnp.github.io/sp-dev-fx-controls-react/) to see if your question is already addressed there. This will help us ensure our documentation is up to date.**

#### Category
[ ] Enhancement

[ ] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [    3.8.0"     ]

#### Expected / Desired Behavior / Question
Expected behavior: When Hovered on the persona the card details should pop up.

#### Observed Behavior
When hovered nothing is happening. Earlier it was working as expected but suddenly it stopped working.
#### Steps to Reproduce

`"@microsoft/sp-core-library": "1.13.1",
    "@microsoft/sp-lodash-subset": "1.13.1",
    "@microsoft/sp-office-ui-fabric-core": "1.13.1",
    "@microsoft/sp-property-pane": "1.13.1",
    "@microsoft/sp-webpart-base": "1.13.1",
    "@pnp/common": "^2.8.0",
    "@pnp/logging": "^2.8.0",
    "@pnp/sp": "^2.13.0",
    "@pnp/spfx-controls-react": "^3.8.0",
    "@pnp/spfx-property-controls": "^3.2.0",
    "bootstrap": "^5.1.1",
    "office-ui-fabric-react": "7.174.1",`

Thanks!

divosuplente:
I confirm that the component is not working. We use a custom component based on the library's LivePersona component. It suddenly stopped working like 2 days ago. No LiveCard is shown when hovering over user coins. No errors are shown in console.

Chipui-Kasar:
I started facing this issue yesterday. Let's wait for their update, hope they'll fix it ASAP.

divosuplente:
Actually, I think there's nothing the team can do to fix it. The control uses a Sharepoint component for it, and my guess is that it is that component that has suddenly stopped working. The LiveCard component this library provides is based on it

frankyvc:
+1 I'm having the same issue. Hopefully there is a quick fix, because this is a blocking issue for my client :( .

AZNak:
+1 same issue here

Vighough:
[...]

PR #1245, opened 2022-06-23
Fixed LivePersona not showing card on hover

SharePickle:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  |  https://github.com/pnp/sp-dev-fx-controls-react/issues/1241

`LivePersona` component stopped working. Microsoft changed `LivePersonaCard` component's property `upn` to `legacyUpn`.

Fixed `LivePersona` component by adding `legacyUpn` property. 
Decided to keep `upn` property in case this is still used on Standard Release tenants.

### DIFF
--- a/src/controls/LivePersona/LivePersona.tsx
+++ b/src/controls/LivePersona/LivePersona.tsx
@@ -37,7 +37,8 @@ if (isComponentLoaded) {
         hostAppPersonaInfo: {
             PersonaType: 'User'
         },
-        upn:  upn,
+        upn: upn,
+        legacyUpn: upn,
         serviceScope: serviceScope,
     }, createElement("div",{},template));
 }


joelfmrodrigues:
@SharePickle do we need both properties? Should the new one not replace the old one? (sorry was unable to test)
image

SharePickle:
> @SharePickle do we need both properties? Should the new one not replace the old one? (sorry was unable to test) image

I would leave both in case MS rolls it back or when your web part is used on Standard Release tenants, which may have previous `LivePersona` component version.

src/controls/LivePersona/LivePersona.tsx
```diff
@@ -37,7 +37,8 @@ if (isComponentLoaded) {
         hostAppPersonaInfo: {
             PersonaType: 'User'
         },
-        upn:  upn,
+        upn: upn,
```
SharePickle, line 40: Maybe we should leave both properties to support previous LivePersona versions? In this case this PR will not break existing components that are deployed on Standard Release tenants.

PR #1259, opened 2022-07-10
Added the possibility to override field rendering for individual field

### LINKED ISSUE #1257: Dynamic Form: add children to override field rendering
#### Category

 [x] Enhancement
 [ ] Bug
 [ ] Question

#### Version

Please specify what version of the library you are using: [3.8.1]

Currently, the DynamicForm does not allow for overriding individual fields. So you get the form as is. If you want one field to behave differently, you have to write the entire form yourself. I would like to suggest that you can add children to the DynamicForm for the fields that you want to render yourself. Let's say you have a list with three fields: Title, SomeTextField and SomeOtherField and you want to do custom rendering for SomeOtherField, the you could put in a DynamicForm like this:

```typescript
<DynamicForm 
          context={this.props.context} 
          listId={"3071c058-549f-461d-9d73-8b9a52049a80"}  
          listItemId={1}
          onCancelled={() => { console.log('Cancelled') }}
          onBeforeSubmit={async (listItem) => { return true; }}
          onSubmitError={(listItem, error) => { alert(error.message); }}
          onSubmitted={async (listItemData) => { console.log(listItemData); }}>
          <SomeCustomFieldComponent internalFieldName="SomeOtherField" /> //Other Properties can be added as well
</DynamicForm>
```

In the render function of the DynamicForm, you can check in the map whether the children of the component contains a child with the specified internalFieldName, if so, that child (in this case SomeCustomFieldComponent) is rendered, if not, nothing changes and the DynamicField is rendered.
This need to be worked out in more detail, but I'd be happy to check if this can be implemented properly and to submit a pull request for this.

bjdekker:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [x]
| New sample?      | [ ]
| Related issues?  | fixes #1257 

Added the possibility to configure that some of the form's fields have a custom renderer. While I initially intended to do this via the children,  I came to the conclusion this was not a very good solution, so I implemented it through the DynamicForm's properties. I introduced the optional fieldOverrides property that can be set as follows:

```typescript
<DynamicForm context={this.props.context} listId={"1d503fec-3680-40f4-9d92-e175cbc0d33c"} onCancelled={() => { console.log('Cancelled'); }} onSubmitted={async (listItem) => { let itemdata = await listItem.get(); console.log(itemdata["ID"]); }}
  fieldOverrides={{"TestField1": (fieldProperties: IDynamicFieldProps) => <DropdownForTextField {...fieldProperties}/> }}
/>
```

This configures that TestField1 should be rendered using the custom DropdownForTextField component, which is a simple implementation of a dropdownfield, that will set the selectedtext as the new value for the field. All other logic (saving etc.) is used from the DynamicForm

```typescript
import { Dropdown, IDropdownOption } from 'office-ui-fabric-react';
import * as React from 'react';
import { IDynamicFieldProps } from '../../../controls/dynamicForm/dynamicField';
export class DropdownForTextField extends React.Component<IDynamicFieldProps, {}> {

    constructor(props: any) {
        super(props);
    }

    public componentDidUpdate() {
    }
    private _onChange = (event: React.FormEvent, option: IDropdownOption, index: number)=> {
        this.props.onChanged(this.props.columnInternalName, option.text);
    }

    public render(): JSX.Element {
        try {
            return (
                <div>
                    The field {this.props.columnInternalName} has a custom renderer <br />
                    <Dropdown onChange={this._onChange}
                        options={[
                            { key: 'A', text: 'Option a' },
                            { key: 'B', text: 'Option b' },
                            { key: 'C', text: 'Option c' },
                            { key: 'D', text: 'Option d' },
                            { key: 'E', text: 'Option e' },
                        ]}/>
                </div>
            );
        } catch (error) {
            console.log(`Error in DynamicField render`, error);
            return null;
        }
    }
}
```

Without the custom renderer specified, everything is rendered according to the default specification (as a TextField in this example  
image

But with this custom renderer, this field is rendered as a dropdown:  
image

image

This will need to be documented in more detail in the documentation, but let's first agree if this is the best way to implement this

### DIFF
--- a/src/controls/dynamicForm/DynamicForm.tsx
+++ b/src/controls/dynamicForm/DynamicForm.tsx
@@ -68,11 +68,18 @@ export class DynamicForm extends React.Component<IDynamicFormProps, IDynamicForm
       fieldCollection,
       isSaving
     } = this.state;
+    
+    const fieldOverrides = this.props.fieldOverrides;
+    
     return (
       <div>
         {fieldCollection.length === 0 ? <div><ProgressIndicator label={strings.DynamicFormLoading} description={strings.DynamicFormPleaseWait} /></div> :
           <div>
             {fieldCollection.map((v, i) => {
+              if(fieldOverrides?.hasOwnProperty(v.columnInternalName)) {
+                v.disabled = v.disabled || isSaving;
+                return fieldOverrides[v.columnInternalName](v);
+              }
               return <DynamicField {...v} disabled={v.disabled || isSaving} />;
             })}
             {
--- a/src/controls/dynamicForm/IDynamicFormProps.ts
+++ b/src/controls/dynamicForm/IDynamicFormProps.ts
@@ -1,5 +1,7 @@
 import { BaseComponentContext } from '@microsoft/sp-component-base';
 import { IItem } from '@pnp/sp/items';
+import React from 'react';
+import { IDynamicFieldProps } from './dynamicField';
 
 export interface IDynamicFormProps {
   /**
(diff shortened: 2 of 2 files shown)


AJIXuMuK:
Thank you @bjdekker for the improvement!

martinlingstuyl:
Nice @bjdekker!

PR #1279, opened 2022-08-10
feat(modern taxonomy picker): ability to disallow selecting children

JakeStanger:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [x]
| New sample?      | [ ]
| Related issues?  | 

Adds new `allowSelectingChildren` prop, which is true by default. Setting to `false` means only the top-level terms appear in search results, and their tree nodes are not expandable. Respects `anchorTermId`.

This is useful in situations where you need people to drill down across multiple fields when paired with `anchorTermId`.

Also includes related docs update.


joelfmrodrigues:
@JakeStanger many thanks for the PR. Could you please resolve conflicts?
I have just updated the target branch from `main` to `dev`.

JakeStanger:
Hey, sorry for the delay - disappeared on holiday & this ended up lost at the bottom of my pile. All resolved :)

joelfmrodrigues (approved):

joelfmrodrigues:
@JakeStanger many thanks for this, great new feature! Now merged and will be available on the beta version within a few minutes.

Just two small notes: 
-in the future, please try to avoid changing existing code if you are only going to change formatting as this adds to the amount of code that we need to review :) Even if it's empty spaces or breaking the code into separate lines, we still need to carefully verify if something changed
-please also avoid updating existing code to generate the same result in a different way, unless it's significantly simpler. In this case, I am referring to the searchTermUrl changes. I personally like the changes you did, but the outcome seems to be exactly the same so ends up being just personal preference and the original code author probably had different preference and we should respect that :) Again, if it's something that will greatly simplify a complex block of code, it's welcome (hard to decide really 😆)

Thank you!

JakeStanger:
Most of those changes came from refactoring as I was expecting to add more complexity but didn't need to, but noted for the future, cheers.

PR #1293, opened 2022-08-24
fix(people picker): fixes #1292 where people picker returns no results.

### LINKED ISSUE #1292: People Picker returns no results on 3.10.0
#### Category
[ ] Enhancement

[x ] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [    3.10.0     ]

Using tenant search in people picker should return results.

#### Observed Behavior
No results are returned and network error is observed.
image

I pulled the latest version of project and debugged and it looks like the issue is with this change:
 
peoplepicker3

I've reverted just those 3 changes back and the people picker is working properly again. I will create a PR with this change if it helps. 

#### Steps to Reproduce
Add the people picker control to a page and do not add groupName or groupId params.

c-eiser13:
| Q               | A
| --------------- | ---
| Bug fix?        | [ x]
| New feature?    | [ ]
| New sample?      | [ ]

Reverted changes previously made for 1.15 support that cause the people picker to return 0 results.

People picker queryParams was including a SharePointGroupID param with null value by default and the API does not allow null values for this param.

### DIFF
--- a/src/services/PeopleSearchService.ts
+++ b/src/services/PeopleSearchService.ts
@@ -110,7 +110,7 @@ export default class SPPeopleSearchService {
     try {
       // If the running env is SharePoint, loads from the peoplepicker web service
       const userRequestUrl: string = `${siteUrl || this.context.pageContext.web.absoluteUrl}/_api/SP.UI.ApplicationPages.ClientPeoplePickerWebServiceInterface.clientPeoplePickerSearchUser`;
-      const searchBody = {
+      const searchBody: any = {
         queryParams: {
           AllowEmailAddresses: true,
           AllowMultipleEntities: false,
@@ -124,12 +124,12 @@ export default class SPPeopleSearchService {
 
       // Search on the local site when "0"
       if (siteUrl) {
-        searchBody.queryParams["SharePointGroupID"] = 0;
+        searchBody.queryParams.SharePointGroupID = 0;
       }
 
       // Check if users need to be searched in a specific SharePoint Group
       if (groupId && typeof (groupId) === 'number') {
-        searchBody.queryParams["SharePointGroupID"] = groupId;
+        searchBody.queryParams.SharePointGroupID = groupId;
       }
 
       // Check if users need to be searched in a specific Microsoft 365 Group, Security Group (incl. nested groups) or Distribution List
--- a/src/services/PeopleSearchService.ts
+++ b/src/services/PeopleSearchService.ts
@@ -119,18 +119,17 @@ export default class SPPeopleSearchService {
           PrincipalSource: 15,
           PrincipalType: this.getSumOfPrincipalTypes(principalTypes),
           QueryString: query,
-          SharePointGroupID: null
         }
       };
 
       // Search on the local site when "0"
       if (siteUrl) {
-        searchBody.queryParams.SharePointGroupID = 0;
+        searchBody.queryParams["SharePointGroupID"] = 0;
       }
 
       // Check if users need to be searched in a specific SharePoint Group
       if (groupId && typeof (groupId) === 'number') {
-        searchBody.queryParams.SharePointGroupID = groupId;
+        searchBody.queryParams["SharePointGroupID"] = groupId;
       }
 
       // Check if users need to be searched in a specific Microsoft 365 Group, Security Group (incl. nested groups) or Distribution List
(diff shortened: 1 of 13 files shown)


SmarterJB:
Hey!
Sadly, this fix destroys again the correct loading of users in the dynamic-form and peoplepicker if you enable all principle types!
`principalTypes={[PrincipalType.User, PrincipalType.SharePointGroup, PrincipalType.SecurityGroup, PrincipalType.DistributionList]}`

It loads a group instead of the correct user!

Cheers,
Josef

c-eiser13:
@SmarterJB hey, curious if the dynamic form works properly in 3.9.0? Based on what I can see in the commit history, the change I made simply reverted the people picker back to what was published in 3.9.0. Wondering if the issue is more with the dynamic form than the people picker? I have not used the dynamic form, but if I get a chance this week, I can try and debug to see what it may be doing.

SmarterJB:
Hey! 

Sorry, it seems that your change has nothing to do with the behaviour it shows! Its just that because the user was not found, the dynamic form used the Title, so it seemed that it works!

But the problem is somewhere else in the DynamicForm using the PeoplePicker when allowing all principalTypes! I dug into it today a little bit, and the issue is that the DynamicForm does not use the e-mail adress of the user in the `defaultSelectedUsers` property, but it uses the id and the title like `{id}/{title}`. But then the peopleSearchService uses the ID as the query, in my case simply "12", and I have a group in my tenant where the UPN starts with a 12, so this gets found instead!

So... Sorry, my comment was wrong, so your PR does not break anything, but the DynamicForm itself combined with PeoplePicker is broken still! I tried today to fix it, but it has a lot of dependencies for saving the DynamicForm if you change the way you store the userdata, so this would take a bit of time! So I think I should open another issue there!

SmarterJB:
So this is how it should be, it should be the user "Josef Benda" selected!
image
This works if the field is set to
image

but if I change the setting to:
image
and hit F5 in my workbench, suddenly the value of the field is different!
image

joelfmrodrigues:
@SmarterJB thanks for the additional information. It would be great if you could open a separate issue please as it really seems like a different one.

src/services/PeopleSearchService.ts
```diff
@@ -119,18 +119,17 @@ export default class SPPeopleSearchService {
           PrincipalSource: 15,
           PrincipalType: this.getSumOfPrincipalTypes(principalTypes),
           QueryString: query,
-          SharePointGroupID: null
```
joelfmrodrigues, line 122: @c-eiser13 was this line causing issues, or are you just removing to revert to the previous working version?

joelfmrodrigues: Ignore...just checked again the details of the related issue and I can see that this was indeed causing issues.

src/services/PeopleSearchService.ts
```diff
@@ -119,18 +119,17 @@ export default class SPPeopleSearchService {
...
       if (siteUrl) {
-        searchBody.queryParams.SharePointGroupID = 0;
+        searchBody.queryParams["SharePointGroupID"] = 0;
       }
 
       // Check if users need to be searched in a specific SharePoint Group
       if (groupId && typeof (groupId) === 'number') {
-        searchBody.queryParams.SharePointGroupID = groupId;
+        searchBody.queryParams["SharePointGroupID"] = groupId;
       }
```
joelfmrodrigues, line 133: Can you revert to using dot notation, please? I assume this was not related to the issue itself?

c-eiser13: @joelfmrodrigues  in the previous version (3.9.0) it was using the [] syntax. Since SharePointGroupID is not defined on the queryParams object on line 114, the dot syntax throw an error on 127 and 132. The issue here is that you cannot pass a null value for SharePointGroupID, so we cannot include it on the queryParams object by default, which is why I believe the [] syntax is being used on 127 and 132. I'm happy to update the PR, just let me know what you think about the above.  Thanks!

joelfmrodrigues: Sorry, you are correct as this would throw an error :) : https://stackoverflow.com/questions/12710905/how-do-i-dynamically-assign-properties-to-an-object-in-typescript

Any chance you could add `any` as the type for searchBody as this should allow the dot notation to be used? If it's too much hassle, let me know and I can update after approving the PR
Many thanks a sorry for so many comments

c-eiser13: @joelfmrodrigues not a problem, just made the changes and added a new commit. Let me know if anything else is needed. Thanks!

joelfmrodrigues:
@c-eiser13 the PR was now including all the previous changes from other PRs as pending changes - not 100% sure why GitHub did not recognise them as existing changes...
I have applied the changes directly to the dev branch and added you as a contributor for the next release as this was just simpler for both of us :)

Many thanks for looking into this and providing a fix!

PR #1295, opened 2022-08-25
Placeholder example to only display in edit mode. Closes #1280 

### LINKED ISSUE #1280: Update docs for `Placeholder` with an example to hide it in read mode
#### Category

- [x] Enhancement
- [ ] Bug
- [ ] Question

#### Version

Please specify what version of the library you are using: Latest

### Update

The feature request is a bit to verbose to implement within the control itself. To compensate, we will update the docs with an example how this can be achieved. 

``` typescript
{
  this.displayMode === DisplayMode.Edit ?
  <Placeholder ...placeholderProps /> :
  <div />  
}
```

More info: https://github.com/pnp/sp-dev-fx-controls-react/pull/1295#discussion_r971673826

> #### Question
> 
> Extending the control `Placeholder` with an extra option that would hide the control when the display mode of the page is in `Read`. This way users could use the control and configure it so the look and feel is similar like a default SP WebPart. For example: When you use the conversations WebPart from Yammer you get the following UI in edit mode.
> 
> afbeelding
> 
> But if you don't configure the WebPart and publish the page, the placeholder is hidden.
> 
> afbeelding
> 
> #### Implementation
> 
> This could be achieved by adding a new variable to the interface `IPlaceholderProps` that looks a bit like the following.
> 
> ``` typescript
> /**
>  * Specify if you want to hide the placeholder when the page is in view mode
>  */
> hideViewMode?: boolean;
> ```

Jwaegebaert:
| Q               | A
| --------------- | :---:
| Bug fix?        | 
| New feature?    | ✔️
| New sample?      | 
| Related issues?  | fixes #1280 

Extending the control Placeholder with an extra option `hideInReadMode` that would hide the control when the display mode of the page is in `read`.

### DIFF
--- a/docs/documentation/docs/controls/Placeholder.md
+++ b/docs/documentation/docs/controls/Placeholder.md
@@ -56,6 +56,21 @@ Sample of using the `hideButton` functionality for hiding the button when page i
              theme={this.props.themeVariant} />
 ```
 
+Sample to only display `Placeholder` when the web part is in edit mode:
+
+```TypeScript
+{
+  this.displayMode === DisplayMode.Edit ?
+  <Placeholder iconName='Edit'
+               iconText='Configure your web part'
+               description='Please configure the web part.'
+               buttonLabel='Configure'
+               onConfigure={this._onConfigure}
+               theme={this.props.themeVariant} /> :
+  <div />
+}
+```
+
 ## Implementation
 
 The placeholder control can be configured with the following properties:


src/controls/placeholder/PlaceholderComponent.tsx
```diff
@@ -120,9 +120,10 @@ export class Placeholder extends React.Component<IPlaceholderProps, IPlaceholder
           const iconTextClassNames = `${styles.placeholderText} ${(this.state.width && this.state.width <= 380) ? styles.hide : ""}`;
           const iconTextEl = typeof iconText === 'string' ? <span className={iconTextClassNames}>{this.props.iconText}</span> : iconText(iconTextClassNames);
           const descriptionEl = typeof description === 'string' ? <span className={styles.placeholderDescriptionText}>{this.props.description}</span> : description(styles.placeholderDescriptionText);
+          const editMode = document.location.href.toLocaleLowerCase().indexOf('mode=edit') !== -1;
```
joelfmrodrigues, line 123: Hi @Jwaegebaert, have you looked at something like this to check if the page is in edit mode?
https://sharepoint.handsontek.net/2018/01/07/detect-sharepoint-environment-type-and-page-mode-from-an-spfx-web-part/

src/controls/placeholder/PlaceholderComponent.tsx
```diff
@@ -120,9 +120,10 @@ export class Placeholder extends React.Component<IPlaceholderProps, IPlaceholder
...
 
           return (
-            <div className={`${styles.placeholder} ${this.props.contentClassName ? this.props.contentClassName : ''}`} ref={this._linkElm}>
+            <div className={`${styles.placeholder} ${this.props.contentClassName ? this.props.contentClassName : ''} ${this.props.hideInReadMode && !editMode ? styles.hide : ''}`} ref={this._linkElm}>
```
joelfmrodrigues, line 126: instead of hiding the element and all the children, can we just return an empty element instead if in read mode? Not sure if this would work, but if it does, seems like a cleaner approach as we avoid adding elements to the page that we know will be always hidden

joelfmrodrigues (commented):
Hi @Jwaegebaert many thanks for the changes, definitely a good improvement. 
I added two comments, could you please review them?

src/controls/placeholder/PlaceholderComponent.tsx
Jwaegebaert: Good suggestion, I've implemented the environment check but left out the one that validates the `displayMode`. Mainly, because when we want to use the `displayMode` validation in the control we would require it as an extra option from the users. Currently, I've left it out and sticked with the following check:
``` javascript 
document.location.href.toLocaleLowerCase().indexOf('mode=edit') !== -1
```

What are your toughts about it?

joelfmrodrigues: @Jwaegebaert  sorry not sure what you mean by: "when we want to use the displayMode validation in the control we would require it as an extra option from the users.". 
What do you mean by require it as an extra option from the users?

My main concern here was to use the URL parameter when the framework does give us that state as a supported property. If MS decides to change the URL parameter, they would break your code

Jwaegebaert: @joelfmrodrigues pardon me for the vague explanation. What I mean with that the user needs to pass along an extra option is that the placeholder control can't access `DisplayMode` because it is a property of `BaseClientSideWebPart`. To access `DisplayMode` we would have to change the property `hideInReadMode` within the interface `IPlaceholderProps` to something like this: 

``` JavaScript
  /**
   * Specify if you want to hide the placeholder when the page is in read mode
   */
  hideInReadMode?: boolean | DisplayMode;
```

This way we could use it to validate the page mode on a modern SharePoint page. At the same time, when the user is using the control on a On Premise SharePoint WebPart they don't have to pass along the DisplayMode and can just use the property as a boolean. 

Few examples how it would look in different project. 
Modern SharePoint WebPart:

``` JavaScript
<Placeholder 
  description='Please configure the web part.' 
  iconName='Edit'  
  iconText='Configure your web part'
  hideInReadMode={this.displayMode} />
```

On Premise SharePoint WebPart:

``` JavaScript
<Placeholder 
  description='Please configure the web part.' 
  iconName='Edit'  
  iconText='Configure your web part'
  hideInReadMode />
```

joelfmrodrigues: @Jwaegebaert Thanks I got it now. But you made me look at it from a different point.
What if, instead of `hideInReadMode` you just add a `hidden` property and let the logic to hide it stay outside of the component (this.displayMode === DisplayMode.Read)? This way the property could be used with any custom scenario and it's a lot more flexible.
Your case to hide it in display mode is possibly a common scenario, so we could add that as an example in the documentation. What do you think?

I know the same could be achieved by adding a condition around the element, so sorry if I am making your original idea too generic, just seemed like it would allow many more scenarios

joelfmrodrigues: @Jwaegebaert just a quick update: I asked the other maintainers (Alex and Fabio) to also look at this PR and share their thoughts regarding the 'hideInReadMode' and 'hidden' as I don't want to change your initial idea unless it makes sense to others as well :)

joelfmrodrigues: @Jwaegebaert please ignore my comments regarding a 'hidden' property...this was completely changing the scope of your PR...

Can you please just make the change to use the DisplayMode property as this would be more reliable than the Url?

Sorry for all the confusion

Jwaegebaert: @joelfmrodrigues those are valid remarks though. Now I'm also thinking that this option could be a bit verbose. If somebody wants to only display `Placeholder` during edit mode they could easily implement it like this:

``` typescript
{
  this.displayMode === DisplayMode.Edit ?
  <Placeholder ...placeholderProps /> :
  <div />  
}
```

Same goes when they want to implement it with a On Premise SharePoint WebPart.

Maybe we should just add an example within the docs to show users how they can achieve this.
What do you think?

joelfmrodrigues:
It's a tricky scenario :) but like you mentioned, it can also be easily done from outside the control (a little bit more work for on-prem, but not a lot).
Would you be willing to submit a PR to update the docs?

src/controls/placeholder/PlaceholderComponent.tsx
joelfmrodrigues: @Jwaegebaert It's a tricky scenario :) but like you mentioned, it can also be easily done from outside the control (a little bit more work for on-prem, but not a lot).
Would you be willing to submit a PR to update the docs?

Jwaegebaert: Will do! 😄

Jwaegebaert:
@joelfmrodrigues the PR has been modified with the docs sample. 😄

joelfmrodrigues:
@Jwaegebaert many thanks!
Merged

PR #1308, opened 2022-09-09
Bugfix/#1306 fix initial term resolution looping

### LINKED ISSUE #1306: Can't use ModernTaxonomyPicker in a controlled way
#### Category
[ ] Enhancement

[x] Bug

[x] Question

#### Version

Please specify what version of the library you are using: [   3.10.0    ] (worked fine with 3.9.0)

It's a common pattern to "control" the value of fields by using some kind of state and define values from the state, then update the state from a onchange

The fluent UI lib for example is full of such pattern (https://developer.microsoft.com/en-us/fluentui#/controls/web/peoplepicker#implementation)

It should possible to reproduce same pattern with ModernTaxonomyPicker.

#### Observed Behavior
If setting initialValues from something that came from the state, the control goes into a infinite loop (which ends in being throttled)
image

If setting initialValues from a prop, whenever the control change, it resets to the initial values.
chrome-capture-2022-8-7

#### Steps to Reproduce
Repro : https://github.com/stevebeauge/pnpreactrepromoderntaxonomy/blob/51969c31f270b51ac5ea4332e232d6a7aa27aa1d/src/webparts/repro/components/Repro.tsx#L1-L53

````typescript
import * as React from 'react';


import { IModernTaxonomyPickerProps, ModernTaxonomyPicker } from '@pnp/spfx-controls-react';

import { BaseComponentContext } from '@microsoft/sp-component-base';
import { TextField } from 'office-ui-fabric-react';
import { useState } from 'react';

type ReproProps = {
  context: BaseComponentContext
};

type PartialTermInfo = IModernTaxonomyPickerProps["initialValues"];

const Repro = ({ context }: ReproProps): React.ReactElement<ReproProps> => {

  const [termSetId, setTermSetId] = useState<string | undefined>(undefined);
  const [selectedTerms, setSelectedTerms] = useState<PartialTermInfo | undefined>(undefined);

  return (
    <>
      <div>
        <TextField value={termSetId} onChange={(evt, newValue) => setTermSetId(newValue)} />
      </div>
      <div>
        {termSetId && (
          <ModernTaxonomyPicker
            termSetId={termSetId}
            panelTitle="My term set"
            label="Modern taxonomy picker"
            context={context}
            initialValues={selectedTerms}
            onChange={setSelectedTerms}
          />
        )}
      </div>
      <div>
        {selectedTerms && (
          <ul>
            {selectedTerms.map(t => (
              <li key={t.id}>{t.labels[0].name}</li>
            ))
            }
          </ul>
        )
        }
      </div>
    </>
  );
}

export { Repro, ReproProps };

````

stevebeauge:
| Q               | A
| --------------- | ---
| Bug fix?        | [X]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #1306

`initialValues` property was part of the dependency of the react effect that resolve the terms. This resolution trigger the onchange event, which results in updating the state in a controlled use of the control. This rerender will set again initial values, and the infinite loop occurs.

Removing this initialValues solve this loop

### DIFF
--- a/src/controls/modernTaxonomyPicker/ModernTaxonomyPicker.tsx
+++ b/src/controls/modernTaxonomyPicker/ModernTaxonomyPicker.tsx
@@ -108,7 +108,9 @@ export function ModernTaxonomyPicker(props: IModernTaxonomyPickerProps): JSX.Ele
           // no-op;
         });
     }
-  }, [currentTermStoreInfo?.defaultLanguageTag, props.anchorTermId, props.context.pageContext, props.initialValues, props.termSetId, taxonomyService]);
+  },
+    // Do NOT add initial values as dependencies because it should trigger term resolution when the component rerender (https://github.com/pnp/sp-dev-fx-controls-react/issues/1306)
+    [currentTermStoreInfo?.defaultLanguageTag, props.anchorTermId, props.context.pageContext, props.termSetId, taxonomyService]);
 
   React.useEffect(() => {
     if (props.onChange && initialLoadComplete.current) {


omarelanis:
This change will break the Taxonomy control as the InitialValues property will no longer work, please see my comment under #1306 for how to correctly use the control

stevebeauge:
I close this PR because I don't think the way I solve it is correct.

There's definitively somthing to look around the Parent attribute of the ItermInfo interface (see discussion in #1306 )

PR #1312, opened 2022-09-14
Use anchor id from taxonomyfield in dynamicform

### LINKED ISSUE #1310: DynamicForm - AnchorId of TaxonomyField gets ignored and the whole tree is rendered
#### Category
[ ] Enhancement

[x ] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: 3.10.0

If I set the anchor in a taxonomy field correctly, the dynamic form should not display all terms from the termset, but instead only the ones from the anchor down, as seen in the screenshot of the standard modern edit form of sharepoint. Here the term "process" was selected to be the anchor of the field.

image

#### Observed Behavior
The dynamic form displays the whole taxonomy tree from the termset as seen in the screenshot below. The term on the top of the tree should be the one named "process" as seen in the screenshot above.

image

#### Steps to Reproduce
Create a document library with a taxonomy field, set the anchor to something lower than the termset, and then open the dynamic form in this library, you can see that it will not use the correct anchor.

I will look into this and will try to create a PR for this issue myself!

Thanks!

SmarterJB:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [x]
| New sample?      | [ ]
| Related issues?  | fixes #1310

This change allows the dynamic form to use the anchor id which is set in the field settings of the taxonomy field. I also changed the wording of the german localization because the word "Laufzeit" in german means "running time" in english! "Knoten" now refers to the "node" of the termstore which is more suiting in my opinion.

This is both a bugfix (because when I used the dynamic form, I wondered why it isnt working) and a new feature!


joelfmrodrigues:
@SmarterJB sorry for the delay on this. Can you please review the conflicts?

SmarterJB:
No problem! I resolved the conflicts!

SmarterJB:
Hello @joelfmrodrigues can this be included in the next release aswell? Then all my current issues should be fixed!

Cheers!

joelfmrodrigues:
Sorry, 3.11.0 was just published some minutes ago (should have been 2 days ago but there was an issue publishing it to npm)
I tried to review the PR before the release but just couldn't find the time.

SmarterJB:
okay no problem! we can work with the beta version once this is through! Thanks for the quick response! :)

src/services/SPTermStorePickerService.ts
```diff
@@ -28,8 +28,6 @@ export default class SPTermStorePickerService {
    * Service constructor
    */
   constructor(private props: ITaxonomyPickerProps, private context: BaseComponentContext) {
-    this.clientServiceUrl = this.context.pageContext.web.absoluteUrl + '/_vti_bin/client.svc/ProcessQuery';
```
AJIXuMuK, line 31: this change is not needed

AJIXuMuK (changes requested):

PR #1316, opened 2022-09-19
Fix - Wrong preselected user in dynamic form

### LINKED ISSUE #1315: DynamicForm - PeoplePicker preselects wrong user if PrincipalType allows groups
Thank you for reporting an issue, suggesting an enhancement, or asking a question. We appreciate your feedback - to help the team understand your
needs please complete the below template to ensure we have the details to help. Thanks!

**Please check out the [documentation](https://pnp.github.io/sp-dev-fx-controls-react/) to see if your question is already addressed there. This will help us ensure our documentation is up to date.**

#### Category
[ ] Enhancement

[x] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [3.10.0]

If the listsetting "Allow selection of" in the "Person or Group"-field is set to "People and Groups", when I open the DynamicForm with a listItemId set, it should display the correct users that are selected in the field.

image

#### Observed Behavior
In the SharePoint-Standard-Modern Edit Form, the selected user is displayed correctly, in this case it is my user in the field "VerteilerInfo":
image

If I want to display this item in the dynamic form, it shows a group:
image

I looked into it, and it seems the problem is, that the dynamic form hands the PeoplePicker component the following `defaultSelectedUsers={["12/Josef Benda"]}`
but in the documentation of the PeoplePicker it says that the component expects user login or email adress instead of the id before the slash!

This causes the search service to find a group, which in my case is a group which email adress starts with "1234PPAP...@....at"!

I think we should rethink the way the default value gets loaded in the DynamicForm, and hand the correct initialValues to the PeopePicker component!

I will try to fix this myself and create a PR, but I think its a lot of rework in the dynamic form!

#### Steps to Reproduce
Create a list with a "People and Groups" field. Then create a list item, and select a user. Create a group which has the id of the user in its email-adress. Then open up the DynamicForm for the list and the item you've created, and you will see that the initial value of the peoplepicker is not correct.

Thanks!

SmarterJB:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #1315 

I changed the way the DynamicForm component fetches the initial data from user fields. I changed it to use the UPN instead of the ID, because as described in issue #1315, this leads to problems when "People and Groups" are enabled in the SharePoint field.

I had to also remove the parsing of the user IDs from the defaultValues, because there are no longer the IDs there.

### DIFF
--- a/src/controls/dynamicForm/DynamicForm.tsx
+++ b/src/controls/dynamicForm/DynamicForm.tsx
@@ -443,7 +443,7 @@ export class DynamicForm extends React.Component<IDynamicFormProps, IDynamicForm
           else if (fieldType === "User") {
             if (item !== null) {
               const userEmails: string[] = [];
-              userEmails.push(await this._spService.getUsersUPNFromFieldValue(listId, listItemId, field.InternalName, this.webURL) + '');
+              userEmails.push(await this._spService.getUserUPNFromFieldValue(listId, listItemId, field.InternalName, this.webURL) + '');
               defaultValue = userEmails;
             }
             else {
--- a/src/controls/dynamicForm/DynamicForm.tsx
+++ b/src/controls/dynamicForm/DynamicForm.tsx
@@ -443,7 +443,7 @@ export class DynamicForm extends React.Component<IDynamicFormProps, IDynamicForm
           else if (fieldType === "User") {
             if (item !== null) {
               const userEmails: string[] = [];
-              userEmails.push(await this._spService.getUserUPNById(parseInt(item[field.InternalName + "Id"])) + '');
+              userEmails.push(await this._spService.getUsersUPNFromFieldValue(listId, listItemId, field.InternalName, this.webURL) + '');
               defaultValue = userEmails;
             }
             else {
--- a/src/controls/dynamicForm/DynamicForm.tsx
+++ b/src/controls/dynamicForm/DynamicForm.tsx
@@ -281,28 +281,16 @@ export class DynamicForm extends React.Component<IDynamicFormProps, IDynamicForm
       field.newValue = [];
       for (let index = 0; index < newValue.length; index++) {
         const element = newValue[index];
-        let retrivedItem = false;
-        if (field.fieldDefaultValue !== null) {
-          if (field.fieldDefaultValue.join(',').indexOf(element.text) !== -1)
-            field.fieldDefaultValue.forEach(item => {
-              if (item.split('/')[1] === element.text) {
-                retrivedItem = true;
-                field.newValue.push(item.split('/')[0]);
-              }
-            });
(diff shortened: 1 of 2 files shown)


SmarterJB:
@joelfmrodrigues maybe someone can look into this? we have a problem with that at a project from us, and this would really help a lot! :) thanks in advance!

joelfmrodrigues:
@SmarterJB  sorry that it took so long to find the time to review this. 
I tested your code and everything seems to work well, except for editing an items with a user field containing a group.
Test list:
image
List item number 1 works fine, and the edit form displays the title and user (Adele) correctly by default.
I then tried with item number 2, and only the title is pre-populated. The group is not loaded.

I then switched back to the dev branch to double-check if this was a gap in your version compared to the current behaviour, and actually got the exact same result. The user is pre-loaded, but not the group.

Can you confirm if these were also the results that you experienced?

This was the part of your comment that made me double-check things:
"I had to also remove the parsing of the user IDs from the defaultValues, because there are no longer the IDs there"
Did we lose any functionality with your changes? (just to ensure I didn't miss something here)

SmarterJB:
Damn @joelfmrodrigues you are right, groups are not loaded correctly, because my select on fieldname/Name for users returns their UPN, but for Groups it just returns the groupname!
```
const apiUrl = `${webAbsoluteUrl}/_api/web/lists(@listId)/items(${listItemId})?@listId=guid'${encodeURIComponent(listId)}'&$select=${fieldName}/Title,${fieldName}/Id,${fieldName}/Name&$expand=${fieldName}`;
```

Will look into this and get you a solution asap!

SmarterJB:
Hey @joelfmrodrigues! I figured it out, it was a problem we had in the code before, because in single user fields there was a request to the sharepoint rest API with `/_api/web/getuserbyid(${userId})` which cannot work with a group as this will always yield a not found exception from the server!

I changed it now so that the value from the single field gets fetched the same way as the values from the multiuser field!

SmarterJB:
But I have no clue why the check failed now!

joelfmrodrigues:
@SmarterJB Many thanks for the super quick update.
I have tested it using the same list but unfortunately, now the form never loads, for both user or group.
My list field is configured as a single value.

Went on debug and the issue seems to be in the foreach loop to parse the response items here:
image

The response I received was not an array:
image

And the following exception was captured
image

Could you please have a look?

SmarterJB:
Hey @joelfmrodrigues! 
My bad! I converted a multiuser field to a singleuser field, thats why it still worked on my test!

I rewrote the single user method from before now it should work!

joelfmrodrigues:
@SmarterJB many thanks! This is now merged and should be available on the beta release in a few minutes.

PR #1335, opened 2022-10-14
Improvements to DynamicForm to support creation of document sets

PaoloPia:
Current implementation of DynamicForm is not working when adding new Document Set instances. This PR fixes the issue and makes it fully functional with Document Set creation.


AJIXuMuK:
Thank you @PaoloPia!

Issue #1337, opened 2022-10-17
Can't use the ContentTypePicker control 


NizarGrindi:
#### Category
[ ] Enhancement

[x] Bug

[x] Question

#### Observed Behavior
The ContentTypePicker control can't be used as described in the [documentation](https://pnp.github.io/sp-dev-fx-controls-react/controls/ContentTypePicker/#implementation)

`(alias) class ContentTypePicker
import ContentTypePicker
JSX element class does not support attributes because it does not have a 'props' property.ts(2607)`

`'ContentTypePicker' cannot be used as a JSX component.
  Its instance type 'ContentTypePicker' is not a valid JSX element.
    Type 'ContentTypePicker' is missing the following properties from type 'ElementClass': context, setState, forceUpdate, props, and 2 more.ts(2786)`

![image](https://user-images.githubusercontent.com/8698368/196144468-e0ccd985-3492-4ed6-84bd-dec872c94262.png)

Thanks

PR #1338, opened 2022-10-17
Fixes#1337 the problem of importing the control

NizarGrindi:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | Can't use the ContentTypePicker control #1337 

The control "ContentTypePicker" can't be used with the preview configuration, it returns an error 

`(alias) class ContentTypePicker
import ContentTypePicker
JSX element class does not support attributes because it does not have a 'props' property.ts(2607)`

`'ContentTypePicker' cannot be used as a JSX component.
  Its instance type 'ContentTypePicker' is not a valid JSX element.
    Type 'ContentTypePicker' is missing the following properties from type 'ElementClass': context, setState, forceUpdate, props, and 2 more.ts(2786)`

[For more details: What is the difference between `import * as react from 'react'` vs `import react from 'react'`](https://stackoverflow.com/questions/54585763/what-is-the-difference-between-import-as-react-from-react-vs-import-react-fr)

### DIFF
--- a/src/controls/contentTypePicker/ContentTypePicker.tsx
+++ b/src/controls/contentTypePicker/ContentTypePicker.tsx
@@ -1,5 +1,5 @@
 import * as telemetry from '../../common/telemetry';
-import React from 'react';
+import * as React from 'react';
 import { IContentTypePickerProps, IContentTypePickerState } from './IContentTypePicker';
 import { ISPService } from '../../services/ISPService';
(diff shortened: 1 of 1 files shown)


joelfmrodrigues:
@NizarGrindi  Many thanks and sorry for the long delay. Merged 🙂

PR #1261, opened 2022-07-11
Features/upgrade to spfx 1.15

stevebeauge:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [X]
| New sample?      | [ ]
| Related issues?  | fixes #X, partially #Y, mentioned in #Z

#### What's in this Pull Request?

Because of compatibility issue with current project, was forced to upgrade to sfpx 1.15 without waiting for an official release.
If it can help the community, here's my progress

PR #1342, opened 2022-10-19
Fix for ModernTaxonomyPicker infinite loop

patrikhellgren:
| Q               | A
| --------------- | ---
| Bug fix?        | [X]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | mentioned in #1306 and #1308

This PR fixes an issue introduced in #1261. The initialValues property was never meant to be used as a way to use the control in a controlled way. The control would need to be totally refactored to make it controlled. In that case another property would preferably be introduced, e.g. selectedItems. See for example the FluentUI [PeoplePicker](https://developer.microsoft.com/en-us/fluentui#/controls/web/peoplepicker) that has both a defaultSelectedItems property to use the control in an uncontrolled way and a selectedItems property to use the control in a controlled way.


AJIXuMuK:
Thanks @patrikhellgren for the expert look at the component!

PR #1343, opened 2022-10-19
Fix term path display in ModernTaxonomyPicker

patrikhellgren:
| Q               | A
| --------------- | ---
| Bug fix?        | [X]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | 

Fixes the display of the term path in ModernTaxonomyPicker so that it is correctly displayed in all places so that it looks more like the out-of-box control.


AJIXuMuK:
Thank you @patrikhellgren for the improvement!

Issue #1346, opened 2022-10-31
SitePicker: add HubId to filter to only sites within a hub


russgove:
#### Category
[x ] Enhancement

[ ] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [  3.7  ]

#### Expected / Desired Behavior / Question
We should add an optional  HubId to the sitePicker Control. If you want to search just for a site that is a member of the hub.

kadu-jr:
Looks like a fair addition. I try to implement it and get back to you over the weekend.

NishkalankBezawada:
Hello @russgove & @kadu-jr 

We already have a property called "[additionalQuery](https://pnp.github.io/sp-dev-fx-controls-react/controls/SitePicker/#implementation)," which, if supplied, will be added as part of the AND operator in the search query. If mode is set to site or web, then it applies.

To obtain the sites that are linked to hub sites, simply enter the hubsite id in the property to the fetch the sites linked to a specific hubsite function.

Below is the sample query text
```ts
(DepartmentId:{Hub.Id} OR DepartmentId:Hub.Id) AND ContentClass:STS_Site AND SiteId<>{Hub.Id} 
```
So, in our case the query would be as below
```ts
(DepartmentId:{f9f37a78-a7f2-4d4d-a294-c5609559d664} OR DepartmentId:f9f37a78-a7f2-4d4d-a294-c5609559d664) AND SiteId<>{f9f37a78-a7f2-4d4d-a294-c5609559d664}
```
and the control code will be as below, 

```tsx
                <SitePicker
                  context={this.props.context}
                  label={'select sites'}
                  mode={'site'}
                  allowSearch={true}
                  multiSelect={false}
                  onChange={(sites) => { console.log(sites); }}
                  placeholder={'Select sites'}
                  searchPlaceholder={'Filter sites'} 
                  additionalQuery={'(DepartmentId:{f9f37a78-a7f2-4d4d-a294-c5609559d664} OR DepartmentId:f9f37a78-a7f2-4d4d-a294-c5609559d664) AND SiteId<>{f9f37a78-a7f2-4d4d-a294-c5609559d664}'}
                  />
```

Let me know if this helps. 

Thanks,
Nishkalank Bezawada

kadu-jr:
Hi @NishkalankBezawada,

[...]

PR #1357, opened 2022-11-14
Added 'associatedsites' and 'hubsiteId' parameters to SitePicker to have option to show only sites associated with the hub. + infinite loop prevention when fetching sites

kadu-jr:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [ X]
| New sample?      | [ ]
| Related issues?  | mentioned in #1346 

Addition of 'associatedsites' option to the 'mode' parameter to search for all sites within a hub, along with a 'hubsiteId' parameter if the user wants to specify the hub look for - otherwise, the current site's hub will be used.

I also added a potential bugfix to prevent an infinite loop when fetching the sites, as this problem was occurring to me.


joelfmrodrigues:
@kadu-jr  many thanks! This has now been merged

PR #1365, opened 2022-11-17
Update ListView.tsx _fieldRender - link validation

### LINKED ISSUE #1171: react list view - is there a way to not make link if the value in linkPropertyName is empty or ''
#### Category
[ ] Enhancement

[ ] Bug

[ x ] Question

#### Version
"@pnp/spfx-controls-react": "2.0.0",
I am on an older version of spfx so I don't want to try to upgrade the version at this point.

I have a view where I'm passing in the 'linkPropertyName' property which is a calculated field which builds up a url.
However, in some cases, the url string will be empty.

I am wondering if there's a simple way (outside of a custom render function), that the list component does not turn it into a link unless it has a valid value (starts with http... or /sites/).

Or does a future version already address this?

#### Observed Behavior

At least in this version, it shows as a link and just goes to the current page when you click on it.

#### Steps to Reproduce

  {
    "name": "LinkTitle",
    "displayName": "Name",
    "minWidth": 150,
    "maxWidth": 300,
    "linkPropertyName": "CalculatedLinkColumn"
  },

Thanks for any tips :)

mikezimm:
| Q               | A
| --------------- | ---
| Bug fix?        | [ x]
| New feature?    | [ x]
| New sample?      | [ ]
| Related issues?  | fixes #1171

## Address this issue:   Empty link Url creates a clickable link
https://github.com/pnp/sp-dev-fx-controls-react/issues/1171

Add basic logic to test if Url property is a valid link structure.

I only tested the raw js in the compiled ListView.js file in my own project.
I reverse engineered the tsx which was pretty basic.
The validation logic could be prettier regex but this long-hand syntax does work and is easier to read.

But if someone could improve it further and validate there are no issues with the update, it would be great!

```typescript
    // Check if the URL property is specified
    if (field.linkPropertyName) {
      return (item: any, index?: number, column?: IColumn) => { // eslint-disable-line @typescript-eslint/no-explicit-any
        
        // Testing to see if Url value is valid... has a value, is a string, and either starts with http or /sites/
        const isValid = item[field.linkPropertyName] && typeof item[field.linkPropertyName] === "string" &&
            ( item[field.linkPropertyName].indexOf("/sites/") === 0 || item[field.linkPropertyName].indexOf("http") === 0 ) ? true : false;

        // Return element as a link if the Url passes simple validation.  Else just return the displayed value as normal span
        return isValid === true ? <a href={item[field.linkPropertyName]}>{item[column.fieldName]}</a> :
          <span>{item[column.fieldName]}</span>;
      };
```


src/controls/listView/ListView.tsx
```diff
@@ -386,7 +386,14 @@ export class ListView extends React.Component<IListViewProps, IListViewState> {
     // Check if the URL property is specified
     if (field.linkPropertyName) {
       return (item: any, index?: number, column?: IColumn) => { // eslint-disable-line @typescript-eslint/no-explicit-any
-        return <a href={item[field.linkPropertyName]}>{item[column.fieldName]}</a>;
+        
+        // Testing to see if Url value is valid... has a value, is a string, and either starts with http or /sites/
+        const isValid = item[field.linkPropertyName] && typeof item[field.linkPropertyName] === "string" &&
+            ( item[field.linkPropertyName].indexOf("/sites/") === 0 || item[field.linkPropertyName].indexOf("http") === 0 ) ? true : false;
```
joelfmrodrigues, line 392: @mikezimm sorry for the delay on this.
The /sites/ validation will fail for example if the site is for example the root site of the tenant or is under /teams/.
Maybe check if the hyperlink string containst the site relative url? 
As for http, I think you can go even further with "https://", what do you think?

joelfmrodrigues:
Hi @mikezimm , just checking if you need any help with the PR?

src/controls/listView/ListView.tsx
mikezimm: @joelfmrodrigues , One other note I forgot to mention:

For the validation check of /sites/:

I am checking to see if the url starts with /sites/ because we can pass in text columns for the url.
In those cases, we sometimes just start the url with /sites/ to keep it shorter.
But if it's the full Url it will always start with http....

If it's a normal Url column, column validation forces proper validation in the first place.

Hope that explains the logic.

mikezimm: > Hi @mikezimm , just checking if you need any help with the PR?

I think I just manually edited the file.  Did I do it wrong or did I miss  a step?

joelfmrodrigues:
@mikezimm just looking into this in more detail and I believe the control already allows you to apply your validation without requiring changes.

This is the current code to render fields:
image

If a `render` property is provided, the component will not return a hyperlink, but instead, use the render function specified for that field. In your case, all you should be required to do is to recreate the same logic on the render method.
If you take as an example the render method highlighted below that is available on the ControlsTest file of this repository, you can just change it to include the validation that you require and return either a hyperlink or text.

image

Is it OK to close this PR as no longer required, or am I missing something?

joelfmrodrigues:
Closing as no response for a long period. Can reopen if required

PR #1371, opened 2022-11-24
Feature/listitempicker support listname

### LINKED ISSUE #1355: ListItemPicker - use list name as well as GUID to point to list
#### Category
[X] Enhancement

[ ] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [  v3.11  ]

Currently, you need the list GUID to point to a list using the 'listId' property. However, can you also add the option to point to a list name? It would be consistent with other PnP patterns.

If we migrate a solution, create a new list with the same name, or reuse a custom component, we need to make sure to change this property based on the new GUID. 

#### Observed Behavior
Add a new property for listName or modify current property to allow entering the list name instead of the GUID.

Thank you!

ekapic:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [x]
| New sample?      | [ ]
| Related issues?  | fixes #1355 

If listId string parameter isn't a Guid, it's treated as a list title and a list guid is pulled from the SP API.


joelfmrodrigues:
@ekapic many thanks for the contribution and apologies for the delay.
This has now been merged

PR #1446, opened 2023-02-06
Search query must query also the upn as mail could be null

eduardpaul:
PeopleSearch service should also find people by userPrincipalName when group transitive membership check is used. If the user does not have an email or the email is different from userPrincipalName currently it is not returned. 

| Q               | A
| --------------- | ---
| Bug fix?        | [ X ]
| New feature?    | [ ]
| New sample?      | [ ]

This PR adds the userPrincipalName filter to the query.


joaojmendes:
@eduardpaul , Thank you for the improvement .

PR #1454, opened 2023-02-13
initialize changedValue with defaultValue

SamueleFurnari:
| Q               | A
| --------------- | ---
| Bug fix?        | [X]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | N.A.

If the field is required, the control shows a required alert with an error signaling even if the field has a value. The PR fix this by picking value from _defaultValue_ for each field.


joaojmendes:
@SamueleFurnari , thank you for the fix.

PR #1455, opened 2023-02-13
Fix image path

SamueleFurnari:
| Q               | A
| --------------- | ---
| Bug fix?        | [X]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | N.A.

Fixs the initial value for an image field by deleting the segment from webUrl, because this is already present in serverRelativeUrl


joaojmendes:
@SamueleFurnari , thank you for the fix.

PR #1456, opened 2023-02-13
check empty array and trasform it in set as null

SamueleFurnari:
| Q               | A
| --------------- | ---
| Bug fix?        | [X]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | N.A.

Fixes an error raised by SharePoint update API when updating an item that contains a field that requires an array. In this UC SharePoint requires a null value instead of an empty array.


joaojmendes:
@SamueleFurnari , Thank you for the fix.

PR #1457, opened 2023-02-13
Fix site browser resize

SamueleFurnari:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | N.A.

Once initialized, the DocumentLibraryBrowser component starts loading the list collection and displays a _Spinner_ control. After receiving the response from SharePoint, the React state was updated successfully but the browser doesn't recalculate the size of the items and so the list is displayed if you don't do a window resize. I redesigned the rendering logic of the component.


joaojmendes:
@SamueleFurnari , Thank you for the fix

PR #1459, opened 2023-02-15
Fix - Show ModernTaxonomyPicker label in correct form

### LINKED ISSUE #1372: Modern Taxonomy Picker confusing label: alloMultipleSelection to false but label says you can select one or more tags
#### Category
[ ] Enhancement

[X] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [ 3.12.0  ]

In Modern Taxonomy Picker or Taxonomy Picker component, when I select "allowMultipleSelections" param to false, in order to allow only one term selection, panel label says "Select one **or more** tags". I think this label should say "Select one tag" for single term selection and "Select one or more tags" for multiple selection. In fact, looking in the location files in other languages -in example, spanish location es-es -, we can see same label ("ModernTaxonomyPickerTreeTitle" key in location files) is translated as "Select one tag", refering only to single selection term. It's a bit confusing for basic users.

#### Observed Behavior
In Modern Taxonomy Picker or Taxonomy Picker, if you choose allowMultipleSelections param as false, at open panel to select a term, the panel label says "Select one or more tags" despite of you only can pick a single term.

#### Steps to Reproduce
In Modern Taxonomy Picker,  set allowMultipleSelections param to false, and open panel to select a term. You can see label Im refering there.

Thanks!

patrikhellgren:
| Q               | A
| --------------- | ---
| Bug fix?        | [X]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #1372

Fixes so that the ```ModernTaxonomyPicker``` label shows in correct form depending on multi or single select option.


joaojmendes:
@patrikhellgren , Thank you for your fix.

PR #1479, opened 2023-02-28
Fixes DynamicForm trying to load TaxonomyFields with wrong display name. Closes #1422

### LINKED ISSUE #1422: Dynamic form trying to load Image tags with wrong internal name
#### Category
[ ] Enhancement

[x] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [3.12.0]

The form to open.

#### Observed Behavior
When the dynamic form is loading, I get a 400 bad request error. This is resulting in a spinner that keeps loading, no feedback to the user.

In developer tools (F12):
**Request:**
/_api/web/lists(@listId)/Fields/getByInternalNameOrTitle('**MediaServiceImageTags_0**')/InternalName?@listId=guid'_thelistguid_'
**Response:**
`{"error":{"code":"-2147024809, System.ArgumentException","message":"Column 'MediaServiceImageTags_0' does not exist. It may have been deleted by another user."}}`

If I inspect my list (SP Insider), the field with internal name _MediaServiceImageTags_0_ does not exist. However, _MediaServiceImageTags_ does. Where do the fields get read from? As a work-around, is there a way to exclude fields from being added to the dynamic form?

#### Steps to Reproduce
Create a subsite in a site collection
Add a document library
Create a dynamic form with react

martinlingstuyl:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #1422

Fixes an issue with Taxonomy Fields concerning the relation between Taxonomy Fields and their corresponding note fields. The corresponding note field internal name was retrieved by assuming that the Display Name of that note field was equal to the internal name of the Taxonomy field and `_0`.

**So for example:**
If the taxonomy field has an internal fieldname called `Project`, the display name of the note field was inferred to be `Project_0`. This is actually a correct assumption in most cases. When creating a Column using the SharePoint UI for example.
But there are scenario's when this is incorrect. For example when fields are provisioned using automated solutions, it's possible for these field titles to have received different values.

It's better to retrieve the field by Id. And this ID is also present in the details of the field in the form of the `TextField` property.

### DIFF
--- a/src/controls/dynamicForm/DynamicForm.tsx
+++ b/src/controls/dynamicForm/DynamicForm.tsx
@@ -426,7 +426,7 @@ export class DynamicForm extends React.Component<IDynamicFormProps, IDynamicForm
               defaultValue = [];
             }
           } else if (fieldType === "TaxonomyFieldTypeMulti") {
-            const response = await this._spService.getTaxonomyFieldInternalName(this.props.listId, field.InternalName, this.webURL);
+            const response = await this._spService.getTaxonomyFieldInternalName(this.props.listId, field.TextField, this.webURL);
             hiddenName = response.value;
             termSetId = field.TermSetId;
             anchorId = field.AnchorId;
--- a/src/services/SPService.ts
+++ b/src/services/SPService.ts
@@ -578,10 +578,10 @@ export default class SPService implements ISPService {
     }
   }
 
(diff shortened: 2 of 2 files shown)


joelfmrodrigues:
@martinlingstuyl many thanks! Merged

martinlingstuyl:
Nice!

PR #1480, opened 2023-03-01
Adds onListItemLoaded handler to DynamicForm. Closes #1472

### LINKED ISSUE #1472: An extra event: `onListItemLoaded` could be useful with Dynamic Forms
#### Category
[x] Enhancement
[ ] Bug
[ ] Question

#### Version
3.11 (Because 3.12 has an issue on dynamicforms)

I'm using Dynamic Forms in combination with fieldOverrides to create custom field experiences. The current setup works when using it in normal situations: you have a simple field and override it, and update the value in the form. 
But if you have multiple fields that have to interact with each other, you need the `state` of the containing component to build it up.
In this situation it becomes necessary (eg when using an edit form) to have a moment when you can fill that state with the values of the listitem.
To do this an extra event is necessary: `onListItemLoaded`. something like this:

```typescript
const fieldOverrides: {[key: string]: any} = {
      "SomeFieldX": (fieldProperties: IDynamicFieldProps) => this.renderSomeFieldX(fieldProperties),
      "SomeFieldY": (fieldProperties: IDynamicFieldProps) => this.renderSomeFieldY(fieldProperties)
};

return <>
<DynamicForm 
    context={this.props.context as any} 
    onListItemloaded={(listItemDate: any)=> { /* do something to update the state */}
    onBeforeSubmit={this._onBeforeSubmit}
    fieldOverrides={fieldOverrides} 
/></>;

```

martinlingstuyl:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [x]
| New sample?      | [ ]
| Related issues?  | Fixes #1472

Adds an extra event handler to the Dynamic Form control. This allows to be able to access list item information after it is retrieved by the DynamicForm when running in Edit and View modes. You could use this to add list item information to the state of the parent component so that you can use it for other purposes. It comes in especially handy when overriding multiple fields that depend on each other using the `fieldOverrides` option of the DynamicForm.

### DIFF
--- a/docs/documentation/docs/controls/DynamicForm.md
+++ b/docs/documentation/docs/controls/DynamicForm.md
@@ -38,6 +38,7 @@ The `DynamicForm` can be configured with the following properties:
 | contentTypeId | string | no | content type ID |
 | disabled | boolean | no | Allows form to be disabled. Default value is `false`|
 | disabledFields | string[] | no | InternalName of fields that should be disabled. Default value is `false`|
+| onListItemLoaded | (listItemData: any) => Promise&lt;void&gt; | no | List item loaded handler. Allows to access list item information after it's loaded.|
 | onBeforeSubmit | (listItemData: any) => Promise&lt;boolean&gt; | no | Before submit handler. Allows to modify the object to be submitted or cancel the submission. To cancel, return `true`.|
 | onSubmitted | (listItemData: any, listItem?: IItem) => void | no | Method that returns listItem data JSON object and PnPJS list item instance (`IItem`). |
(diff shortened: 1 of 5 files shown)


joaojmendes:
@martinlingstuyl , thank you very much for the fix.

martinlingstuyl:
> @joelfmrodrigues, I did have time left over to fix this. Question, what am I do do with the extra file-updates that are auto generated after my commit?
> 
> image
> 
> 
> 
> Should I add it to the PR?

What should I do in general with these files @joaojmendes?

PR #1483, opened 2023-03-03
fix: file picker recent tab wont render until page re-sized, adds def…

### LINKED ISSUE #1482: FilePicker - recent tab empty until re-render
#### Category
[ ] Enhancement

[ x] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [  1.13.0       ]

When the file picker opens and defaults to the recent tab, should show results.

#### Observed Behavior
When the file picker opens and defaults to the recent tab, nothing is shown until a re-render. I can resize the page, or navigate to a different tab, stock images for example, then back to Recent and content is shown correctly. Initial load:
image

After resizing the page and a re-render:
image

In debugging the project, looks like the issue is in the _getItemCountForPage method when trying to set the column width, the surfaceRect.width prop is 0, so the width is not set correctly until a re-render.
FilePickerNaNError

I've added some default values for columnCount and columnWidth when the surfaceRect.width is 0 and this solves the issue, I just do not work with the Fluent List control much so not sure this is the best solution.
image

I will create a PR with this change for review.

#### Steps to Reproduce
I was able to debug the project and repro by testing in the ControlsTest web part that is part of the project. Open the file picker when defaulting to the Recent tab and notice there are no results.

c-eiser13:
…ault values to avoid this issue.

| Q               | A
| --------------- | ---
| Bug fix?        | [ x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #1482 

This PR adds some default values for columnWidth and columnCount when the passed in surfaceRect.width is 0 to avoid the NaN issue that causes the recent tab to be blank on initial load.

See #1482 for more information on this issue.

### DIFF
--- a/src/controls/filePicker/RecentFilesTab/RecentFilesTab.tsx
+++ b/src/controls/filePicker/RecentFilesTab/RecentFilesTab.tsx
@@ -41,11 +41,10 @@ export default class RecentFilesTab extends React.Component<IRecentFilesTabProps
 
     this._selection = null;
 
-
     this.state = {
       isLoading: true,
       results: [],
-      filePickerResults: []
+      filePickerResults: [],
     };
   }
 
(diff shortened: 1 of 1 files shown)


joaojmendes:
@c-eiser13 , Thank you for the fix!

Issue #1484, opened 2023-03-03
`ListItemAttachments` : Cannot download items when it has a ilegal character


6gal6ler6:
# Category
[ ] Enhancement

[x] Bug

[ ] Question

#### Version

[3.13.0](https://pnp.github.io/sp-dev-fx-controls-react/about/release-notes/#3130)

## Bug Found

When trying to open a file on the `ListItemAttachments` control, it won’t download if the File name has a name that is URL reserved.

## Steps to reproduce

- Use the `ListItemAttachments` control
- Add a bunch of files with names like
   - File # 1.msg
   - File #  2.xml
   - File # 3.docx
- Try to download any file using `ListItemAttachments` control
- Resulting URL: https://TENANT.sharepoint.com/sites/SITE/Lists/LISTTITLE/Attachments/ITEMID/File  **the rest of the name is gone**

PR #1485, opened 2023-03-03
`ListItemAttachments` : FIX: Cannot download items when it has a ilegal character 

6gal6ler6:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]

Encode URI to fix the issue of file download when it has URL reserved characters such as "#"

Reported Issue #1484


joaojmendes:
@6gal6ler6 , Thank  you very much for the fix !

PR #1490, opened 2023-03-14
DynamicForm - Fixing Required Multi Field Saving Problem

### LINKED ISSUE #1489: Dynamic Form allows submitting when required Fields with Multiple-Selection get removed
#### Category
[ ] Enhancement

[ x ] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: v3.13.0

If I remove the value of a required field which allows multiple selections, the Dynamic Form should display an error message at the required field and not allow saving the form.

#### Observed Behavior
Currently you can just delete the values of a required field which allows multiple selections and it will not show the missing value and save the empty value to the list item.

I have already fixed the problem, and a PR is on the way!

#### Steps to Reproduce
Create a list with a field that allows multiple selections (e.g. Multi Select Field) and make it required. Now you can add a value to the field and then open the Dynamic Form, and you be able to remove the value from the required field and then save the list item, without getting any error message or response about the required field.

Cheers!

SmarterJB:
| Q               | A
| --------------- | ---
| Bug fix?        | [ x ]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #1489 

Changed the way the required fields are validated in the DynamicForm.tsx, and also how the rendering of the error message is handled in the DynamicField.tsx. In both files it was expected that an empty value consists out of null, undefined or an empty string, but not an empty array. This PR fixes this problem

### DIFF
--- a/src/controls/dynamicForm/DynamicForm.tsx
+++ b/src/controls/dynamicForm/DynamicForm.tsx
@@ -104,7 +104,7 @@ export class DynamicForm extends React.Component<IDynamicFormProps, IDynamicForm
   }
 
   //trigger when the user submits the form.
-  private onSubmitClick = async (): Promise<void> => {
+    private onSubmitClick = async (): Promise<void> => {
     const {
       listId,
       listItemId,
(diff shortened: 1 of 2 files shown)


joaojmendes:
@SmarterJB  thank you very much for the fix.

PR #1500, opened 2023-04-03
Changed react import to fix "cannot be used as a JSX component" error

andreasomayrat:
| Q               | A
| --------------- | ---
| Bug fix?        | [ x ]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  |

Changed React import in FieldPicker to "import * as React from 'react' " to prevent " 'FieldPicker' cannot be used as a JSX component"-error that appears in some projects

### DIFF
--- a/src/controls/fieldPicker/FieldPicker.tsx
+++ b/src/controls/fieldPicker/FieldPicker.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import * as React from 'react';
 import cloneDeep from 'lodash/cloneDeep';
 import { Dropdown, IDropdownOption, IDropdownProps } from 'office-ui-fabric-react/lib/Dropdown';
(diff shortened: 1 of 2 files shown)


fabiofranzini:
Hi @andreasomayrat thanks for this fix!
I encountered a problem in the Web Part to test the control... You used a List Id that is obviously not present on my Tenant (same for other users). This causes the Web Part to fail causing it not to render. 
Can you fix this?
Could you for example use the 'ListPicker' control for the demo and pass the List Id to the 'FieldPicker' control?
Thank you :)

andreasomayrat:
Hello @fabiofranzini 

I have now updated the code to use the selected list from the ListPicker component

joelfmrodrigues:
@andreasomayrat sorry for the delay, this is now merged. Many thanks for your contribution

PR #1530, opened 2023-05-03
Guidozam patch dynamicfield

GuidoZam:
| Q               | A
| --------------- | ---
| Bug fix?        | [ x ]
| New feature?    | [ ]
| New sample?      | [ ]

Fix to the DynamicField control that prevented the correct behavior of the getRequiredErrorText method when the changedValue is null.

### DIFF
--- a/src/controls/dynamicForm/DynamicForm.module.scss
+++ b/src/controls/dynamicForm/DynamicForm.module.scss
@@ -46,7 +46,7 @@
 	margin-right: 8px;
 }
 
-.feildDisplayNoPadding {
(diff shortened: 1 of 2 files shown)


joelfmrodrigues:
@GuidoZam thanks for the PR. I just merged the other PR with the fix to the property name and the same changes are showing here. Could you please update this PR?

GuidoZam:
Sorry @joelfmrodrigues my bad, I thought I've created another branch for the changes. I'll fix as soon as I can!

PR #1559, opened 2023-06-16
 fix issue with initial rendering of the TilesList 

### LINKED ISSUE #1558: FilePicker - Tile view issue on first render
#### Category
[ ] Enhancement

[ X] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [ 1.15.0 ]

"Tiles" view should use all available space for rendering the items on first render

#### Observed Behavior
On initial rendering "Tiles" view option shows its items in a single column with fixed width of 232px. 

FilePicker-initial-tile-view-rendering

After selecting an item or perform window resizing - it renders properly.
FilePicker-tile-view-after-item-selection

I think the problem comes from these default values (lines 100-102):
https://github.com/pnp/sp-dev-fx-controls-react/blob/dev/src/controls/filePicker/controls/TilesList/TilesList.tsx#L98

Without setting the container width on this line, it works correctly - https://github.com/pnp/sp-dev-fx-controls-react/blob/dev/src/controls/filePicker/controls/TilesList/TilesList.tsx#L150

#### Steps to Reproduce
Open FilePicker and navigate to a folder. Make sure "View option" is set to "Tiles". It will render the elements in a single fixed-width column.

DMichev:
| Q               | A
| --------------- | ---
| Bug fix?        | [X]
| New feature?    | [ ]
| New sample?      | [ ]

Resolves #1558 issue

### DIFF
--- a/src/controls/filePicker/controls/TilesList/TilesList.tsx
+++ b/src/controls/filePicker/controls/TilesList/TilesList.tsx
@@ -53,8 +53,14 @@ export class TilesList extends React.Component<ITilesListProps> {
   }
 
   public componentDidUpdate(prevProps: ITilesListProps): void {
-    if (this.props.filePickerResults !== prevProps.filePickerResults) {
+    if ((this.props.filePickerResults !== prevProps.filePickerResults) || (this._pageWidth === -1)) {
       this._listElem.forceUpdate();
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


joelfmrodrigues:
@DMichev many thanks for this. Tested and everything seemed to work as expected

PR #1576, opened 2023-06-30
Fix ModernTaxonomyPicker used language by checking for supported term…

### LINKED ISSUE #1573: ModernTaxonomyPicker can't find term when UI is in language not supported by term store
#### Category
[ ] Enhancement

[X] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [3.13.0]

Using the ModernTaxonomyPicker when the SharePoint UI is in a language not supported by the term store (= not in termStoreInfo.languageTags), we should be able to search terms from the text field and find them in the default term store language.

#### Observed Behavior
Using the ModernTaxonomyPicker when the SharePoint UI is in a language not supported by the term store (= not in termStoreInfo.languageTags), while searching terms from the text field, no term can be found at all (the call to the termStore unofficial SP API is returning a 500 error code with the following details: "Specified argument was out of the range of valid values.\r\nParameter name: lcid").

moderntaxonomypicker lang error

#### Steps to Reproduce
- Use the ModernTaxonomyPicker with another UI language than english from a SharePoint site where the site collection termstore only supports english language (or with any other language not supported by the termstore)
- Search for any term from the text input
- No result will be return

#### Personal note
Looking at the code, I think it should not be hard to implement a fix.
After the first render, the language used by the component is set like this:
```
const languageTag = props.context.pageContext.cultureInfo.currentUICultureName !== '' ?
          props.context.pageContext.cultureInfo.currentUICultureName :
          currentTermStoreInfo.defaultLanguageTag;
```
As we already have the `currentTermStoreInfo`, we could just check the `currentUICultureName` against the `termStoreInfo.languageTags` array and use `currentTermStoreInfo.defaultLanguageTag` if the current UI language is not supported (= not inside the `termStoreInfo.languageTags` array) by the term store.

Thanks!

Zhephyr54:
… store languages

| Q               | A
| --------------- | ---
| Bug fix?        | [ X]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #1573

Changing how the language used by the ModernTaxonomyPicker is set by checking the currentUICultureName against the termStoreInfo.languageTags array and use currentTermStoreInfo.defaultLanguageTag if the current UI language is not supported (= not inside the termStoreInfo.languageTags array) by the term store.
I also noticed that the `currentTermStoreInfo` state was used directly after being set (it would be undefined because state update is asynchronous) which must have caused another bug if the currentUICultureName wasn't available. So I used the fetched `termStoreInfo` instead of the state. 

Thanks!

### DIFF
--- a/src/controls/modernTaxonomyPicker/ModernTaxonomyPicker.tsx
+++ b/src/controls/modernTaxonomyPicker/ModernTaxonomyPicker.tsx
@@ -80,9 +80,9 @@ export function ModernTaxonomyPicker(props: IModernTaxonomyPickerProps): JSX.Ele
     taxonomyService.getTermStoreInfo()
       .then((termStoreInfo) => {
         setCurrentTermStoreInfo(termStoreInfo);
-        const languageTag = props.context.pageContext.cultureInfo.currentUICultureName !== '' ?
+        const languageTag = props.context.pageContext.cultureInfo.currentUICultureName !== '' && termStoreInfo.languageTags.includes(props.context.pageContext.cultureInfo.currentUICultureName) ?
           props.context.pageContext.cultureInfo.currentUICultureName :
-          currentTermStoreInfo.defaultLanguageTag;
+          termStoreInfo.defaultLanguageTag;
         setCurrentLanguageTag(languageTag);
         setSelectedOptions(Array.isArray(props.initialValues) ?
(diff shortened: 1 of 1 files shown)


joelfmrodrigues:
@Zhephyr54 many thanks, this is now merged

PR #1585, opened 2023-07-11
DynamicForm Number min max

GuidoZam:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [ ]
| New sample?      | [ ]
|Enhancement      | [ x]

Enhanced the check for minimum and maximum number values in DynamicForm control. The control now supports also fields that requires only the minimum or maximum limitation and not always both of them.

I translated all the localization strings with Google translate, hopefully those are correct 🤞


GuidoZam:
This is what it looks like with the PR:

image

joaojmendes:
@GuidoZam Thank you very much for the PR

Issue #1604, opened 2023-07-31
[DynamicForm] - Number validation is preventing form save in certain circumstances, not enabled for currency fields


t0mgerman:
#### Category
[ ] Enhancement

[x ] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [ v3.15 ]

#### Expected / Desired Behavior / Question
I've identified two issues:

Firstly, number validation added in 8998b5dc4b72097f2aa132a4163829d30f52bfc0 seems to have broken form validation where:
- a numeric list field is set as not required 
- that numeric list field has a minimum and maximum value set
- the user chooses not to enter a value, or enters a value then deletes it

I had a some lists where fields had minimum and maximum values between 1 and 5 - and they all stopped working in DynamicForm after recently updating the library in one of my projects.

Secondly, the validation added does not yet work for Currency fields.

#### Observed Behavior
The form silently fails to save if the user leaves the field blank. The reason this happens is because the check in `DynamicForm.tsx` sets `shouldBeReturnBack` to true if `val.newValue < val.minimumValue`. If the user leaves a field empty, `val.newValue` is an empty string `""` - which is falsey, and equivalent to zero. So if the minimum value is, for example, 1 - then `shouldBeReturnBack` is set to true, and the form doesn't save. For a list where the field is not required, this is incorrect behaviour. In `DynamicField.tsx` the logic that determines the error to display DOES check the length of the string value first, and so - no error is returned to tell the user why the form didn't save.

#### Steps to Reproduce
- Add a number column to a list 
- Set the minimum and maximum values to 1 and 5 respectively. 
- Using DynamicForm, enter and then delete a value from this field and attempt to save a form while this value is blank. 

For the currency issue, observe that validation added in the referenced commit does not also work for those fields.

I have a PR ready to go for the above.

PR #1601, opened 2023-07-20
Fix - Issue 1592 - Dynamic form - Number validations are working, but the percentage values are not getting saved

NishkalankBezawada:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #1592 

#### What's in this Pull Request?
Fix for issue #1592 

#### Problem
The number validations are working like a charm, but if we are using a percentage column and trying to save the value to that column, it is not saving the values. Below is the screen for the validations

image

#### Steps to Reproduce
1) Add field Number and choose as 'Show as percentage'
2) Use this list in Dynamic form, with New Item, and give the value as 15 and click save
The form will not be saved. 

#### Solution
The properties number field if chosen as 'Shown as percentage', the minimum and maximum values will be retrieved as below, 

image

wherein, the properties of number field without choosing 'Show as percentage', the minimum and maximum values will be retrieved as below. 

image

And at ```onSubmitClick```, the validation for percentage column is failing due to the minimum and maximum values for the percentage column. 

image

While validating a number, the percentage column is validated as below, 
image

While ```onSubmit``` the same validation for the percentage column is not done, instead, the minimum value and maximum value is checked with the default values as 0.1 and 1 respectively. 

So, a new method is introduced to check the validation again while the user is submitting the form as below, 
image

Thanks,
Nishkalank

PR #1605, opened 2023-07-31
Fix for number field validation and behaviour, addition of currency constraint validation

### LINKED ISSUE #1592: Dynamic Controls - Number validations on the Percentage column is working, but the value in the percentage column is not getting saved.
#### Category
[ ] Enhancement

[x] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [3.15.0 Beta version]

Entering the value between 1 to 100 into percentage column should save the form. 

#### Observed Behavior
The number validations are working like a charm, but if we are using a percentage column and trying to save the value to that column, it is not saving the values. Below is the screen for the validations

image

#### Steps to Reproduce
1) Add field Number and choose as 'Show as percentage'
2) Use this list in Dynamic form, with New Item, and give the value as 15 and click save

The form will not be saved. 

#### Observations
The properties number field if chosen as 'Shown as percentage', the minimum and maximum values will be retrieved as below, 

image

wherein, the properties of number field without choosing 'Show as percentage', the minimum and maximum values will be retrieved as below. 

image

And at ```onSubmitClick```, the validation for percentage column is failing due to the minimum and maximum values for the percentage column. 

image

Thanks,
Nishkalank Bezawada

t0mgerman:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [x]
| New sample?      | [ ]
| Related issues?  | fixes #1592 #1604 

There is some overlap with the PR here - https://github.com/pnp/sp-dev-fx-controls-react/pull/1601#issue-1814203674
I don't believe my branch, with the changes I've made, has the issue referenced in #1592 

This PR makes a few different changes:

### Fixes number validation issues
- In `onSubmitClick` in `DynamicForm.tsx` - current code does some validation to determine whether the form should actually save a list item or not. The Number validation added to better support min and max number settings (in this commit 8998b5dc4b72097f2aa132a4163829d30f52bfc0) added logic here in an `else` statement following `if (val.isRequired)`. This means that even if a field is violating it's own min and max constraints - the form is only preventing save when the field _isn't required_. I have changed this logic to capture constraint violations that occur whether the field is required or not.
- When a number column is set to "show as percentage" - the numbers entered in Dynamic Form were being multiplied by 100. I believe this is because values saved via the REST API are typically expected to be between 0 and 1 for percentages; so 15 was becoming 1500% in the list. This is not how the OOB list form behaves, so I am doing a conditional division by 100 to counter this effect.
- The logic was silently failing the form in cases where users left the field blank and the field had a minimum value constraint greater than 0. This is the issue discussed in #1604 -- I have amended the expressions here so that the form does not fail. Existing code earlier in the loop will catch a scenario where the field is blank when also required. The code I have added will check against constraints only when the input is not empty. It will also fail the form in instances where the input is an invalid number (NaN).

### Fixes native number input behaviour when using keys / scroll
- `<TextField/>` accepts attributes common to `HTMLInputElement` - which is why `type="number"` and `type="password"` among others would work. Pressing the up or down keys or using the mouse wheel while a number field has focus causes the number to go up or down in value. This can be constrained by min and max values if they are provided, but they were not. By pulling them in, we now get correct behaviour - numbers do not go above or below the bounds set in list settings.

### Extending of number validation behaviour to Currency fields, and enhancement of error message to format values as currency
- Number validation has been extended to currency fields, by giving them the same custom error message shared by regular number fields. 
- To display the correct currency symbol in the error message for currency fields, I get the installed LCIDs from SharePoint when DynamicForm loads using `PnP/sp`. Using the added `CurrencyMap.ts` file included in this PR, the LCID set on any currency field can be converted to a culture tag like `en-GB`. If the lookup fails, we use the culture provided in `pageContext` instead. This can then be used with `Intl.NumberFormat` to display values correctly - so if a currency field has a maximum value of 50 - the error message will now say "Value must be lower than £50.00". If we were sure available languages in SPO will remain fairly static, the call to regional settings could be replaced with another mapping object (a `Record<number,string>` object etc).


michaelmaillot:
Hi @t0mgerman,

Thanks for this submission!

Following @NishkalankBezawada's merged PR #1601 that covers #1592, would you mind updating your branch to remove conflicts so that I'll have a look at it?

t0mgerman:
Hi @michaelmaillot - I'll take a look at this tomorrow. 

I've nearly got another branch to publish / PR to submit which refactors parts of `DynamicForm`, `DynamicField` and the `ControlsTestWebPart` - it adds validation expression evaluation and custom formatting (of header, body, footer) to DynamicForm, and I have amended the ControlsTestWebPart so that toggled components and settings persist across reloads. I have a bit of testing to do to make sure I haven't broken anything!

Would it make more sense for me to ensure #1601 is already conflict merged with that new branch, ensure these changes are added and abandon this PR?

t0mgerman:
@michaelmaillot I took a look and merged conflicts. Some of the things @NishkalankBezawada was doing in #1601 in `DynamicForm` I was covering in changes to `DynamicField` - but functionally they achieve the same thing. Tested merged code on a list with Number column, Number as Percentage, and Currency.

I'll raise a different PR for the work I discussed above, as you may wish to give it more of a review.

AJIXuMuK:
Thanks @t0mgerman for the changes!

PR #1648, opened 2023-09-19
[ListItemAttachments] - Upload files in sequence not parallel

### LINKED ISSUE #1644: ListItemAttachments - Inconsistent file handling
Thank you for reporting an issue, suggesting an enhancement, or asking a question. We appreciate your feedback - to help the team understand your
needs please complete the below template to ensure we have the details to help. Thanks!

**Please check out the [documentation](https://pnp.github.io/sp-dev-fx-controls-react/) to see if your question is already addressed there. This will help us ensure our documentation is up to date.**

#### Category
[ ] Enhancement

[x?] Bug

[x] Question

#### Version

Please specify what version of the library you are using: [ 3.15.0  ]

I am experiencing inconsistent uploading and deletion of files. Often, perhaps only when creating a new list item, a few files will often not be uploaded. Checking the dev tools, the addFile-method fails with a 404 or 409.
I don't know if my code is the issue (this is my first webpart, and I am generally not a terribly experienced developer), but I can't figure out what the issue is. Things seem to work fine on an existing item.
Another observation is that when attaching files to an existing list item, I see the loading animation running, but I don't see that when attaching to a new list item; not sure if that says anything.

Regarding deletion, I sometimes experience that a file, often the last after having added a few, does not get deleted but instead remains in the container.

Thank you so much! I've spent days troubleshooting this and really hope I can get help figuring it out!

/Rikke

Tanddant:
As pr. issue #1644 there's an issue when uploading multiple attachments at the same time to the same list item

| Q               | A
| --------------- | ---
| Bug fix?        | [X]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes  #1644

This PR changes the "order" in which attachments are uploaded, before they where uploaded in parallel, however, doing so sometimes causes SharePoint to throw an HTTP 409 when adding several attachments at the same time, this change makes it so the files are uploaded in one long sequence preventing the issue

### DIFF
--- a/src/controls/listItemAttachments/ListItemAttachments.tsx
+++ b/src/controls/listItemAttachments/ListItemAttachments.tsx
@@ -76,12 +76,14 @@ export class ListItemAttachments extends React.Component<IListItemAttachmentsPro
 
   public async uploadAttachments(itemId: number): Promise<void> {
     if (this.state.filesToUpload) {
-      await Promise.all(this.state.filesToUpload.map(file => this._spservice.addAttachment(
-        this.props.listId,
-        itemId,
-        file.name,
-        file,
-        this.props.webUrl)));
+      for (const file of this.state.filesToUpload) {
+        await this._spservice.addAttachment(
+          this.props.listId,
+          itemId,
+          file.name,
+          file,
+          this.props.webUrl);
+      }
     }
     return new Promise<void>((resolve, reject) => {
       this.setState({


joaojmendes:
Hi @Tanddant, thank you for your PR; what do you think we will have the two options for uploading the files?  We can define a property on the control the user can choose, what chance he wants applied. Because the parallel can add some performance to the upload...  

What do you think?
Thank you! again!

Tanddant:
@joaojmendes I agree that there's a performance impact, if we do parallel upload we should add better error handling, I kept seeing this fail, even with two documents - so maybe we should allow it but add a warning about it

joelfmrodrigues:
@Hey Tanddant , when you experience the parallel upload failing, do you see any error messages? What is the volume of files when you experience this? Does it happen with just a few, or when trying to upload a large volume (hundreds)?

Tanddant:
Hi @joelfmrodrigues, the component does not show/throw any errors, but looking in the network history I see the calls are failing with a 409, it seems to be "easiliest" replicated when uploading smaller files which might not finish uploading in order, I just tried to upload 10 txt files that said hello world, and could replicate it every time.

Let me know if you can't replicate, would be happy to provide a video demonstration.

You can also try the sample code i added over in #1644, to see if maybe different tenants behave differently 🤔

joelfmrodrigues:
Likely also solves https://github.com/pnp/sp-dev-fx-controls-react/issues/1682

Tanddant:
Very much looks like it's the same issue, don't know if MS changed something behind in the backend that causes this issue to be more prominent

joelfmrodrigues:
@Tanddant sorry that it took so long to get back to this. It's been a bit chaotic at my side recently.
Merged and will be in the next release

PR #1671, opened 2023-10-06
FilePicker Tiles view render pages

VictorRomanov1986:
| Q               | A
| --------------- | ---
| Bug fix?        | [X ]
| New feature?    | [ ]
| New sample?      | [ ]

This PR fixes an issue, when List rendering more than 1 pages of items and selecting the last item led to scrolling list to the top and displaying only first page.


src/controls/filePicker/controls/TilesList/TilesList.tsx
```diff
@@ -76,6 +76,8 @@ export class TilesList extends React.Component<ITilesListProps> {
               getItemCountForPage={this._getItemCountForPage}
               getPageHeight={this._getPageHeight}
               onRenderPage={(pageProps: IPageProps, defaultRender?: IRenderFunction<IPageProps>) => this._onRenderPage(pageProps, defaultRender)}
+              renderedWindowsAhead={this.props.items.length}
```
AJIXuMuK, line 79: Isn't that too much of extra windows?
If I have 20 items - we will render extra 40 windows.

AJIXuMuK (changes requested):
Thanks @VRomanovTau for the PR!
I have one question in the review - could you please answer/address it?

src/controls/filePicker/controls/TilesList/TilesList.tsx
VRomanovTau: I am not really sure what these properties are responsible for, but without having them in such values component re-renders, when selecting the last item when there is more than 2 pages and displaying only 1st page. With less value it starts to re-render and randomly scrolling to random position

AJIXuMuK:
Sorry @VRomanovTau for the long delay. 
Do you have some test I could you to repro the issue?
I don't think the correct fix is to render so many pages.

michaelmaillot:
Hello @VictorRomanov1986,

Did you have to the chance to post some tests to repro the issue?

michaelmaillot:
Hello @VictorRomanov1986,

What's the status of this PR? Do you need any assistance?

michaelmaillot:
Ok, feel free to open a new one anytime 🙂

Issue #410, opened 2019-11-06
ListView: How to style header?


robiharid:
I want to apply a style to my header row

Inspecting the element shows me the list view has two divs, both with role called 'presentation', one of them is the header with a class of DetailsHeader

how can I style this? Best to add onto the DetailsHeader?

EDIT: Better yet, how can I remove the header so that I can add my own Title? setting displayName: ' ', still declares the space around it

michaelmaillot:
Hi @robiharid,

For now there's nothing available to customize header (only to set it as sticky or not).

But I think that this could be an improvment idea to suggest 🙂 .

Does this answer to your question?

PR #1700, opened 2023-11-09
ListView - Added ability to apply header styling

bradleypregon:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [x]
| New sample?      | [ ]
| Related issues?  | Partially adds #410

(This is my 1st open source contribution and I'm still fresh at TS/JS... bear with me.)

I added the ability to apply custom styles to the header of a ListView via optional prop headerClassName. 

Something I was not able to implement was the ability to modify the font styling of the header text (size, weight, etc), so I set the header min-height to 17px which seemed small enough to scrunch everything without clipping any text - I tried testing as many scenarios as I could. Theoretically, to modify the font styling, I think you could add another styling prop to ListView, and apply that prop to '.ms-DetailsHeader-cellTitle' in ListView.tsx... I can experiment with this at another time.

For _my_ use case, although likely unconventional, I'm using a ListView in a tight spot on a page and need all the real estate being taken up by the default padding (padding-top of 16px) and line height (42px iirc). 

Let me know if I'm, at all, doing this incorrectly (lol).

### DIFF
--- a/src/controls/listView/ListView.tsx
+++ b/src/controls/listView/ListView.tsx
@@ -572,17 +572,40 @@ export class ListView extends React.Component<IListViewProps, IListViewState> {
       return null;
     }
 
+    const headerProps = { 
+      ...props,
+      className: this.props.headerClassName,
+      styles: {
+        root: { 
+          lineHeight: "normal",
+          minHeight: '17px',
+          '.ms-DetailsHeader-cell': {
+            whiteSpace: 'normal',
+            lineHeight: 'normal',
+            height: '100%',
+          },
+          '.ms-DetailsHeader-cellTitle': {
+            height: '100%',
+            display: 'flex',
+            flexDirection: 'row',
+            alignItems: 'center',
+            flexWrap: 'nowrap',
+            justifyContent: 'flex-start',
+          },
+          ...props.styles,
+        },
+      },
+    }
+
     if (this.props.stickyHeader) {
       return (
         <Sticky stickyPosition={StickyPositionType.Header} isScrollSynced>
-          {defaultRender({
-            ...props,
-          })}
+          {defaultRender(headerProps)}
         </Sticky>
       );
     }
 
-    return defaultRender(props);
+    return defaultRender(headerProps)
   }
   /**
    * Default React component render method
--- a/src/controls/listView/IListView.ts
+++ b/src/controls/listView/IListView.ts
@@ -80,6 +80,11 @@ export interface IListViewProps {
    */
   listClassName?: string;
   /**
+   * Class name to apply additional styles on list view header
+   * (min-height 17px)
+   */
+  headerClassName?: string;
+  /**
    * Custom sorting function.
    * @returns sorted collection of items
    */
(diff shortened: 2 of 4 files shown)


bradleypregon:
i have no idea how the other 2 commits got in here. Oops?

src/controls/listView/ListView.tsx
```diff
@@ -572,17 +572,40 @@ export class ListView extends React.Component<IListViewProps, IListViewState> {
...
     }
 
+    const headerProps = { 
+      ...props,
+      className: this.props.headerClassName,
+      styles: {
+        root: { 
+          lineHeight: "normal",
+          minHeight: '17px',
+          '.ms-DetailsHeader-cell': {
```
joelfmrodrigues, line 582: Referencing the class name is likely to fail in the future as we have no control over style updates from external components. Can you think of a more reliable approach?

src/controls/listView/ListView.tsx
```diff
@@ -572,17 +572,40 @@ export class ListView extends React.Component<IListViewProps, IListViewState> {
...
+            whiteSpace: 'normal',
+            lineHeight: 'normal',
+            height: '100%',
+          },
+          '.ms-DetailsHeader-cellTitle': {
```
joelfmrodrigues, line 587: same as above regarding class name

joelfmrodrigues (changes requested):
Hi @bradleypregon, many thanks and sorry for the long delay.
The approach seems fine to me, but I added 2 comments regarding reference to CSS classes by name as that approach often causes issues when the underlying components are updated and classes change. Please let me know if you have any questions

bradleypregon:
Ah, I understand your reasoning about the class names. I'm not sure of a good alternative at the moment, I'll have to do some digging and tinkering.

michaelmaillot:
Hi @bradleypregon,

Are you still working on this PR? Do you need help?

bradleypregon:
Hello @michaelmaillot, 

Apologies for the late reply. Yes, I am still working on the PR, although I have not looked at it for some time, and I'm still stuck. 

I tried a few different approaches, one involving the use of `mergeStyleSets` from the office-ui-fabric-react package, to no avail. If memory serves, every approach I found relied on utilizing the styling class name directly. I'm just not sure of a solid and practical way to access/modify the list header styling.

michaelmaillot:
I had a closer look at both the `ListView` control & the Fluent UI `DetailsList` one. From my perspective, you can't customize individual columns from the header directly (which you referenced through the `.ms-DetailsHeader-cell` and `.ms-DetailsHeader-cellTitle`).

My suggestion would be to keep your PR that focuses on the global header styling, then think of another one for improving the `viewFields` prop, in order to add new property.

Any thoughts regarding this @joelfmrodrigues @AJIXuMuK @joaojmendes @fabiofranzini?

joelfmrodrigues:
@michaelmaillot agree

michaelmaillot:
Hello @bradleypregon,

Any updates regarding this PR?

joaojmendes:
@bradleypregon, any updates on this?  Thanks!

PR #1704, opened 2023-11-13
Use stringMatchOption instead of deprecated stringMatchId #1688

### LINKED ISSUE #1688: ModernTaxonomyPicker not displaying suggestions when typing in values - "API not found" error
**Category**

[ ] Enhancement

[x] Bug

[ ] Question

**Version**
Please specify what version of the library you are using: [3.15.0]

**Expected / Desired Behavior / Question**
when users starts to type into the ModernTaxonomyPicker field they should be presented with a list of suggestions.

**Observed Behavior**

When typing into the ModernTaxonomyPicker the suggestions menu returns "no results found". however there are a number of terms in that area of the term store that should return a match. When i have looked in the console there as a 400 bad request error from the searchTerm api.
The field is anchored correctly because the users can click the slide out and select terms this way. It just fails to display suggestions because the API call fails.

The request URL is formulated as follows in the network console: 

https://####.sharepoint.com/sites/#####/_api/v2.1/termstore/searchTerm(label='a',setId='{guidvalue}',languageTag='en-US',stringMatchId='0',parentTermId='{guidvalue}')?$top=50
 
the response form this call is: 

{
    "error": {
        "code": "invalidRequest",
        "innerError": {
            "code": "apiNotFound"
        },
        "message": "API not found"
    }
} 

I have managed to get the call working in a browser window by changing the perameter stringMatchId='0' to stringMatchOption='StartsWith'. 

**Steps to Reproduce**

1) Add the ModernTaxonomyPicker control in a SPFx project
2) Anchor the picker to a parent term that has children
3) start typing a term value into the picker field that matches one of the child terms.

lafe:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #1688

As discussed in #1688 it appears that the `stringMatchId` query parameter is not supported by the searchTerm-API anymore. The fix proposed in that issue (using the `stringMatchOption` query parameter with value `StartsWith`) worked.

With VS Code I could not find any usages of the `searchTerm` function that uses the `stringMatchId`, so there are (hopefully) no side effects in the package itself. For consuming libraries, this might be a breaking change.

### DIFF
--- a/src/services/SPTaxonomyService.ts
+++ b/src/services/SPTaxonomyService.ts
@@ -58,13 +58,13 @@ export class SPTaxonomyService {
     }
   }
 
-  public async searchTerm(termSetId: Guid, label: string, languageTag: string, parentTermId?: Guid, allowSelectingChildren = true, stringMatchId: string = '0', pageSize: number = 50): Promise<ITermInfo[]> {
+  public async searchTerm(termSetId: Guid, label: string, languageTag: string, parentTermId?: Guid, allowSelectingChildren = true, stringMatchOption: string = 'StartsWith', pageSize: number = 50): Promise<ITermInfo[]> {
     try {
       const query = [
         `label='${label}'`,
         `setId='${termSetId}'`,
         `languageTag='${languageTag}'`,
-        `stringMatchId='${stringMatchId}'`
+        `stringMatchOption='${stringMatchOption}'`
       ];
 
       if(parentTermId !== Guid.empty) {


raclettierer:
Any news on this bug ?

joelfmrodrigues:
@lafe many thanks and apologies for the delay in reviewing this. This is now merged

Issue #1495, opened 2023-03-29
Readme documents highlight extension does not work correctly


ValerasNarbutas:
Thank you for reporting an issue, suggesting an enhancement, or asking a question. We appreciate your feedback - to help the team understand your
needs please complete the below template to ensure we have the details to help. Thanks!

**Please check out the [documentation](https://pnp.github.io/sp-dev-fx-controls-react/) to see if your question is already addressed there. This will help us ensure our documentation is up to date.**

#### Category
[ ] Enhancement

[X] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [         ]
[Latest]
_If you are not using the latest release, please update and see if the issue is resolved before submitting an issue._

#### Expected / Desired Behavior / Question
Looks like were are breaking symbols in  readme files under guides or contributing:
[https://pnp.github.io/sp-dev-fx-controls-react/guides/contributing/](https://pnp.github.io/sp-dev-fx-controls-react/guides/contributing/)

![image](https://user-images.githubusercontent.com/16476453/228536705-f87111d0-30ee-4158-afd0-f16f7b8fe53a.png)

I can work on this one @joaojmendes 
Just wondering is extension used does not support "```sh " and should be "```bash"

change this:
```sh
# comment
```

to this:
```bash
# comment
```

In all documents which using "#" for commenting.

Thanks!

michaelmaillot:
Hi @ValerasNarbutas,

Are you still interested into submitting a PR? 🙂

ValerasNarbutas:
Hey, Sure @michaelmaillot, i can do it

ValerasNarbutas:
i did forget this one :) 
i will check it soon

ValerasNarbutas:
@michaelmaillot added fixes.

PR #1724, opened 2023-12-13
fix bug in readme display issue #1495

ValerasNarbutas:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes ##1495

Having  sh incode does not compile in web version, changed to bash

### DIFF
--- a/docs/documentation/docs/guides/contributing.md
+++ b/docs/documentation/docs/guides/contributing.md
@@ -27,7 +27,7 @@ Typos are embarrassing! Most PR's that fix typos will be accepted immediately. I
 Before contributing:
 
 - ensure that the **dev** branch on your fork is in sync with the original **sp-dev-fx-controls-react** repository
-    ```sh
+    ```bash
     # assuming you are in the folder of your locally cloned fork....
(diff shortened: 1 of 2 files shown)


michaelmaillot:
Hi @ValerasNarbutas,

Thanks for this PR!

May I ask you which version of Material for MkDocs did you use during your tests?

Because even if there's no doubt that we have to upgrade current supported version (3.10), I've tested your changes with this one and the display is still wrong.

But if I remove tabs indent for the code blocks, it's working. Example with the `contributing.md` page:

**Before**

image

image

**After**

image

image

I thinkg that for now, the best fix would be to remove tabs in the code blocks. Then we'll update the version of Material for MkDocs.

Are you ok with this?

ValerasNarbutas:
Yes, off course! 

Not sure about version:) had similar issue in my repo, now thinking maybe I need to recheck it with your solution.

AJIXuMuK:
Thanks you @ValerasNarbutas!

PR #1745, opened 2024-01-24
Added customIcons property on DynamicForm.

GuidoZam:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [x]
| New sample?      | [ ]

Added a new property on the DynamicForm control called __customIcons__. This property is used if the developer wants to override the default icons generated automatically by the DynamicForm.

To use it simply define a dictionary with the internalColumnName as key and set the corresponding Fluent UI icon name.
I've also updated the ControlsTest web part with a sample using the Title column.


AJIXuMuK:
@GuidoZam - could you please resolve the conflicts

GuidoZam:
@AJIXuMuK 👍 I'll do as soon as I can.

GuidoZam:
@AJIXuMuK it should be ok now 😄

PR #1747, opened 2024-01-29
Added orderBy to DynamicField props for lookup fields

GuidoZam:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [x]
| New sample?      | [ ]

Added __orderBy__ property on the __DynamicField__ control to allow sorting when using directly the __DynamicField__ control.

For example if you have a __DynamicForm__ and you want to declare a custom sorting for lookup _TargetField_ you can do the following:

```Typescript
private _customRenderTargetField = (fieldProperties: IDynamicFieldProps): React.ReactElement<IDynamicFieldProps> => {
  return <DynamicField
      key="TargetField"
      {...fieldProperties}
      orderBy={"Title desc"}
      />;
}
```


AJIXuMuK:
Thanks @GuidoZam!

PR #1765, opened 2024-02-14
fix(taxonomy tree): options remain selected when switching in single mode

JakeStanger:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]

This is a hack fix since the tree is already built on top of React anti-patterns, which cause this in the first place. In the future a full refactor is really required to sort this out at its roots.

By adding an unstable `key`, it forces the component to re-render on every render. Without this, React assumes nothing has changed and does not react to the `selectedKey` update.

For large trees with everything expanded, this could cause performance issues but should be okay in most cases.


joelfmrodrigues:
Hi @JakeStanger many thanks for the update. To be completely sure that we don't break things for other developers already using the component that may have a very large collection, and since there is a small risk involved, could you please add a new property to the component to control enabling this behaviour (disabled by default)?

PR #1768, opened 2024-02-17
className and style property added in RichText control

HarminderSethi:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [x]
| New sample?      | [ ]

className property  assigned to ReactQuill which was missing and also added style property. Both will provide the flexiblity to style the control as per the needs.


joelfmrodrigues:
Hi @HarminderSethi many thanks for this. While reviewing, I noticed that the className property is already being added to the top-level element of the render method (line 519)
image

Please add a new property instead of re-using the existing one as I assume that specifying styles for both components could sometimes become tricky.  Of course, I am happy to be convinced otherwise if you believe one property is the best approach.

src/controls/richText/RichText.tsx
```diff
@@ -467,7 +467,7 @@ export class RichText extends React.Component<IRichTextProps, IRichTextState> {
     }
 
     // Okay, we're in edit mode.
-    const { placeholder, styleOptions: { showStyles, showBold, showItalic, showUnderline, showAlign, showList, showLink, showMore, showImage } } = this.props;
+    const { placeholder,className, styleOptions: { showStyles, showBold, showItalic, showUnderline, showAlign, showList, showLink, showMore, showImage } } = this.props;
```
joelfmrodrigues, line 470: Can you please also add props.style here so it follows the same code style?

src/controls/richText/RichText.types.ts
```diff
@@ -14,10 +14,14 @@ export interface IRichTextProps {
...
-   * @defaultvalue null
+   * @defaultvalue undefined
    */
   className?: string;
-
+ /**
+   * Styles to apply to the rich text editor.
+   * @defaultvalue undefined
+   */
+  style?: React.CSSProperties
```
joelfmrodrigues, line 24: Can you please also add the new property to the docs?

joelfmrodrigues (changes requested):

joelfmrodrigues:
Apologies @HarminderSethi , one more thing: I have updated the PR to target the `dev` branch and there is a conflict that needs to be resolved. Can you please do this by pulling the latest version from the dev branch?
It's a few things but hopefully all quick to resolve. Let us know if you need help with any issues.

HarminderSethi:
@joelfmrodrigues  Thanks for reviewing and sharing your inputs. I had to delete my fork repository as I created the fork with master branch only. Created new fork and submitted the new pull request.

Also after analysing the changes I found that className property for React Quill control will not make any impact so removed it. and also style property is added to root div similar to className to make consistentency of the implementation.

Issue #1416, opened 2023-01-07
[Enhancement] Implement custom sorting of fields in a DynamicForm


tahawkins21:
#### Category
[x] Enhancement

[ ] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [Latest]

#### Expected / Desired Behavior / Question

I would like to see the ability to control the order of list columns that are rendered using a DynamicForm component.
Assumption is that a property could be added to IDynamicFormProps.tsx "fieldOrder" that takes an array of internal field names that will be rendered in order, or similar.

It looks like IDynamicFieldProps already has an Order prop to work with, though I haven't figured out how to make that work in DynamicForm.

Thanks!

wuxiaojun514:
It looks like dynamic form uses the column order on content type level , you can configure content type column order as work around
![image](https://user-images.githubusercontent.com/29603048/215381854-95bf64d9-ea30-45fa-8755-0a5de96c280e.png)

michaelmaillot:
Hi @tahawkins21,

Following linked PR merged, this enhancement should be available in the next beta release.

Thanks to you for raising this and @srpmtt for the contribution!

PR #1802, opened 2024-04-06
DynamicForm - Custom sorting

srpmtt:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [ X ]
| New sample?      | [ ]
| Related issues?  | mentioned in #1416 

DynamicForm custom sorting.
Added a new optional prop _fieldOrder_ (string[]), the values are the fields internal names.

### DIFF
--- a/src/controls/dynamicForm/DynamicForm.tsx
+++ b/src/controls/dynamicForm/DynamicForm.tsx
@@ -147,7 +147,12 @@ export class DynamicForm extends React.Component<
    * Default React component render method
    */
   public render(): JSX.Element {
-    const { customFormatting, fieldCollection, hiddenByFormula, infoErrorMessages, isSaving } = this.state;
+    let {fieldCollection} = this.state
+    const { customFormatting, hiddenByFormula, infoErrorMessages, isSaving } = this.state;
+
+    if (this.props.fieldOrder) {
+      fieldCollection = this.sortFields(fieldCollection, this.props.fieldOrder)
+    }
 
     const customFormattingDisabled = this.props.useCustomFormatting === false;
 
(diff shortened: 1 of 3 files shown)


src/controls/dynamicForm/DynamicForm.tsx
```diff
@@ -147,7 +147,12 @@ export class DynamicForm extends React.Component<
    * Default React component render method
    */
   public render(): JSX.Element {
-    const { customFormatting, fieldCollection, hiddenByFormula, infoErrorMessages, isSaving } = this.state;
+    let {fieldCollection} = this.state
+    const { customFormatting, hiddenByFormula, infoErrorMessages, isSaving } = this.state;
+
+    if (this.props.fieldOrder) {
+      fieldCollection = this.sortFields(fieldCollection, this.props.fieldOrder)
```
michaelmaillot, line 154: Is there a reason to sort fields during rendering instead of doing it during mouting / `fieldOrder` prop update?

From my perspective, this could be done in the `getListInformation` method, triggered in mount / update event, to avoid unnecessary re-ordering and which is already doing all the work around list / list item / fields.

src/controls/dynamicForm/IDynamicFormProps.ts
```diff
@@ -122,4 +122,11 @@ export interface IDynamicFormProps {
    * The key is the field internal name and the value is the Fluent UI icon name.
    */
   customIcons?: { [columnInternalName: string]: string };
+
+  /**
+   * Specify fields custom sorting.
+   * The value is the field internal name.
+   */
+  fieldOrder?: string[]
```
michaelmaillot, line 130: Please update the `DynamicForm` doc page to reflect that new feature.

src/controls/dynamicForm/DynamicForm.tsx
```diff
@@ -147,7 +147,12 @@ export class DynamicForm extends React.Component<
    * Default React component render method
    */
   public render(): JSX.Element {
-    const { customFormatting, fieldCollection, hiddenByFormula, infoErrorMessages, isSaving } = this.state;
+    let {fieldCollection} = this.state
+    const { customFormatting, hiddenByFormula, infoErrorMessages, isSaving } = this.state;
+
+    if (this.props.fieldOrder) {
```
michaelmaillot, line 153: It would be better to check the length of the array (as it could be empty if controlled outside for example) to avoid unnecessary Map set.

michaelmaillot (changes requested):
Thanks for this PR @srpmtt!

I've added a few suggestions, let me know if you have any questions.

michaelmaillot (approved):

michaelmaillot:
Merged manually, thank you!

PR #1872, opened 2024-08-21
Fix 1862

### LINKED ISSUE #1862: DynamicForm - Form should hide fields that are hidden on the list content type
#### Category
[ ] Enhancement

[x] Bug

[ ] Question

If I'm rendering a form on a library with content types, and some fields are hidden at list content type level, those fields should not show up in the rendered dynamic form. 

#### Observed Behavior
Fields that are hidden at content type level currently show up in the rendered form, I now have to work around that manually by adding these fields to the `hiddenFields` option, which is a shame, because I need to add quite some custom code to make that work.

#### Steps to Reproduce
Create a list content type with a couple of hidden fields, and render a DynamicForm, using the content type id:

```ts
public render(): React.ReactElement<{}> {
    return <>
      <DynamicForm 
        context={(this.props.context as any)} 
        contentTypeId={this.props.context.contentType.id}      
      />
    </>;
```

NishkalankBezawada:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #1862 

Fixing issue 1862 - Dynamic form should hide fields that are hidden on the List Content Type 

#### Issue 
When a form on a library with content types is rendered, and some fields are hidden at list content type level, those fields are still shown on the form. This is still on the latest release v3.19.0 as below, 

image

#### Solution 
A simple check as below, 

```
      // Process fields that are not marked as hidden
      if (hiddenFields.indexOf(field.InternalName) < 0) {
        if(field.Hidden === false) { *** added new condition 
        *** other logic here***
        }
```

Making ```field.Hidden === false``` resolves the issue as below, 

image

#### Guidance
@martinlingstuyl 

Thanks,
Nishkalank Bezawada

### DIFF
--- a/src/controls/dynamicForm/DynamicForm.tsx
+++ b/src/controls/dynamicForm/DynamicForm.tsx
@@ -1069,265 +1069,267 @@ export class DynamicForm extends React.Component<
 
       // Process fields that are not marked as hidden
       if (hiddenFields.indexOf(field.InternalName) < 0) {
-        order++;
-        let hiddenName = "";
-        let termSetId = "";
-        let anchorId = "";
-        let lookupListId = "";
-        let lookupField = "";
-        const choices: IDropdownOption[] = [];
-        let defaultValue = null;
-        let value = undefined;
-        let stringValue = null;
-        const subPropertyValues: Record<string, unknown> = {};
-        let richText = false;
-        let dateFormat: DateFormat | undefined;
-        let principalType = "";
-        let cultureName: string;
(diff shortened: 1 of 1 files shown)


joaojmendes:
@NishkalankBezawada Thank you for the fix

NishkalankBezawada:
> @NishkalankBezawada Thank you for the fix

Pleasure is mine 😇

PR #1903, opened 2024-11-13
Adds ability to create files/folders in subfolder using DynamicForm. Closes #1901

### LINKED ISSUE #1901: [Dynamic Form] Cannot create folder in library folder
#### Category
[x] Enhancement

[x] Bug

[ ] Question

Not sure if I would call this a bug. It certainly is unexpected behavior though.

#### Version

Please specify what version of the library you are using: 3.18.1

It's not resolved in the latest version 3.19.

I can use the Dynamic Form on a library for creating document sets. However, the document sets will always be created in the root of the library. Even if I've clicked the new button from a subfolder. I'd like to be able to create something in the folder where I currently am located.

#### Observed Behavior
A document set, created using a dynamic form, is always created in the root of the library.

#### Steps to Reproduce
1. Add the document set feature to a sharepoint site / library.
2. Add a dynamic form in your spfx solution, using the contentTypeId of the document set.
3. Navigate to a subfolder in the library. Click new
4. Create the document set using the dynamic form.

#### Location in code

The location  where this goes wrong:
https://github.com/pnp/sp-dev-fx-controls-react/blob/master/src/controls/dynamicForm/DynamicForm.tsx#L590C11-L592C13

```typescript
const newFolder = await library.rootFolder.addSubFolderUsingPath(
  folderTitle
);
```

It should take into account the `RootFolder` of the form, which can be a subfolder of the library. This information is available as a querystring variable.

martinlingstuyl:
Closes #1901

| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [x]
| New sample?      | [ ]
| Related issues?  | Closes #1901

In its current state, a dynamic form always creates new items/files/folders in the root folder of a list or library. This is not very handy when using the dynamic form on document libraries. For example as a custom form to create document sets.

This PR adds the ability to create files, folders and document sets in subfolders. 

I've added an extra option to do this. So the user has full control over the folder that should be used. In this way it doesn't matter how the form is rendered. A `RootFolder` querystring parameter may be available that the user can use. (For example in the NewForm of an SPFx FormCustomizer) Or it may not be available. The user is in control of what value he supplies...

> The feature does not work for list items. Even if list items can also be created in folders. I've not added that currently as it was a little bit more work that I'm not sure is very relevant.

#### Testing instructions

To test this feature, you'll need a library where you can create document sets. In `TestForm.tsx` fill in the `contentTypeId`, `folderPath` AND `hiddenFields` properties as follows:

```tsx
<DynamicForm
          context={this.props.context}
          contentTypeId='0x0120D52000796BC7FCC8B819488729FB2B77B0E799002B2379630DCCB04D883603B763E843D7'
          hiddenFields={['FileLeafRef']}
          folderPath='Some subfolder/Some deeper subfolder'
          onListItemLoaded={async (listItemData: any) => { // eslint-disable-line @typescript-eslint/no-explicit-any
            console.log(listItemData);
          }} />
```

For the `folderPath` option, you can also use a server relative URL. For example when using the RootFolder querystring of an SPFx form customizer. However: it should be a path to the same library, otherwise the code will fail.

`FileLeafRef` should be excluded in `hiddenFields`, otherwise the form will not submit. This is supposed to be fixed in #1906.

### DIFF
--- a/docs/documentation/docs/controls/DynamicForm.md
+++ b/docs/documentation/docs/controls/DynamicForm.md
@@ -66,6 +66,7 @@ The `DynamicForm` can be configured with the following properties:
 | validationErrorDialogProps | IValidationErrorDialogProps | no | Specifies validation error dialog properties |
 | customIcons | { [ columnInternalName: string ]: string } | no | Specifies custom icons for the form. The key of this dictionary is the column internal name, the value is the Fluent UI icon name. | 
 | storeLastActiveTab | boolean | no |  When uploading files: Specifies if last active tab will be stored after the Upload panel has been closed. Note: the value of selected tab is stored in the queryString hash. Default - `true` |
+| folderPath | string | no | Server relative or library relative folder to create the item in. This option is only available for document libraries and works only when the contentTypeId is specified and has a base type of type Document or Folder. Defaults to the root folder of the library. |
 
 ## Validation Error Dialog Properties `IValidationErrorDialogProps`
 | Property | Type | Required | Description |
--- a/src/controls/dynamicForm/DynamicForm.tsx
+++ b/src/controls/dynamicForm/DynamicForm.tsx
@@ -31,6 +31,7 @@ import "@pnp/sp/lists";
 import "@pnp/sp/content-types";
 import "@pnp/sp/folders";
 import "@pnp/sp/items";
+import { IFolder } from "@pnp/sp/folders";
 import { IInstalledLanguageInfo } from "@pnp/sp/presets/all";
 import { cloneDeep, isEqual } from "lodash";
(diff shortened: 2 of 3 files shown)


martinlingstuyl:
@joelfmrodrigues, there it is. Hope its up to the repo standards 😉 We didn't discuss a lot beforehand, so I just took a course that seemed most logical to me. Do let me know if you've got a different opinion!

joelfmrodrigues:
@martinlingstuyl could you please resolve the conflicts?

martinlingstuyl:
@joelfmrodrigues, done!

joelfmrodrigues:
@martinlingstuyl  many thanks, this is now merged

PR #1908, opened 2024-11-19
Fix #1877: WebPartTitle Control Fails to Update After Initial Save

### LINKED ISSUE #1877: WebPartTitle Control Fails to Update After Initial Save
#### Category
[ ] Enhancement

[x] Bug

[ ] Question

#### Version

v3.19.0

Affects also previous versions.

#### Expected Behavior
The WebPart title should be editable every time the WebPart is in edit mode, not just during the initial setup.

#### Observed Behavior
- The WebPart title can only be set once, during the initial addition of the WebPart.
- The event appears to propagate correctly, but the WebPart does not re-render, leading to the observed faulty behavior.
- Note: Users have reported that this faulty behavior started appearing a couple of days ago.
- Forcing a re-render by adding this.render() resolves the issue, but this would require updating all existing components. Additionally, this requirement is not documented as mandatory.

`updateProperty: (value: string) => {
          this.properties.title = value;
          this.render();
        }}
`

### Steps to Reproduce:
1. Scaffold a new SPFx React solution using SPFx v1.19.0.
2. Ensure you are using Node.js v18.19.1.
3. Implement the `WebPartTitle` Control as per the [PnP WebPartTitle documentation](https://pnp.github.io/sp-dev-fx-controls-react/controls/WebPartTitle/).
4. Add the WebPart to a SharePoint page.
5. Publish the page.
6. Edit the page and attempt to update the WebPart title.

EnablingLogic:
| Q               | A
| --------------- | ---
| Bug fix?        | [X]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #1877

This PR fixes the issue described in #1877. The `WebPartTitle` component switched to use `value` in stead of the `defaultValue` property on the `textarea`. This causes the value to not being updated.

### DIFF
--- a/src/controls/webPartTitle/WebPartTitle.tsx
+++ b/src/controls/webPartTitle/WebPartTitle.tsx
@@ -54,7 +54,7 @@ export class WebPartTitle extends React.Component<IWebPartTitleProps, {}> {
           <div className={styles.webPartTitle} style={{ color: color }}>
             {
               this.props.displayMode === DisplayMode.Edit && (
-                <textarea placeholder={this.props.placeholder ? this.props.placeholder : strings.WebPartTitlePlaceholder} aria-label={strings.WebPartTitleLabel} onChange={this._onChange} value={this.props.title} />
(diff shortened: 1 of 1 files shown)


joelfmrodrigues:
@EnablingLogic many thanks for looking into this, fix is now merged

PR #1924, opened 2024-12-17
Fixes #1847 - Multiple selections possible when allowMultipleSelection set to false

### LINKED ISSUE #1847: ModernTaxonomyPicker - allowMultipleSelections = false, but panel allows multiple items to be selected
#### Category
[ ] Enhancement

[X] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: 3.18.1

When allowMultipleSelections is false, I shouldn't be able to select more than one item in the tree view.

#### Observed Behavior

Multiple items can be selected in the tree view in the panel. It still only allows selecting one item in the picker itself, but the treeview should not allow multiple selections.

#### Steps to Reproduce

```
<ModernTaxonomyPicker
                allowMultipleSelections={false}
                termSetId={this.state.businessGroupTermSetId}
                panelTitle="Select Business Group"
                label="Business Group or Unit"
                context={this.props.context}
              />
```
Then, just click the Label button to open the panel and try to select more than one.

image

A change in  `onPickerChange` is most likely required to only select one item when `allowMultipleSelections` is false: https://github.com/pnp/sp-dev-fx-controls-react/blob/master/src/controls/modernTaxonomyPicker/taxonomyPanelContents/TaxonomyPanelContents.tsx#L66

ferrarirosso:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #1847

SelectedKey is correctly retrieved.


joelfmrodrigues:
@ferrarirosso many thanks!

PR #1936, opened 2025-01-19
Dispose of Async instance to prevent memory leak

KevinTCoughlin:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]

Disposes of `Async` instance so it can be garbage collected.


src/controls/sitePicker/SitePicker.tsx
```diff
@@ -67,7 +67,11 @@ const styles = mergeStyleSets({
   },
 });
 
-const async = new Async();
+useEffect(() => {
```
AJIXuMuK, line 70: with this change `async` is not initialized anymore so it will be `undefined`.
Could you please update the code to return back initialization (maybe as a memo or ref)?

Thank you!

AJIXuMuK (changes requested):

src/controls/sitePicker/SitePicker.tsx
KevinTCoughlin: My mistake since this was in eval scope. `useAsync` will memoize and dispose properly within FC scope - https://github.com/microsoft/fluentui/blob/c28f7314c8a967a0f6f7bd6e4a7df9a5ab7703fc/packages/react-hooks/src/useAsync.ts#L7. I updated the branch, thanks.

AJIXuMuK (approved):

AJIXuMuK:
Thank you @KevinTCoughlin for the change!

PR #1954, opened 2025-02-20
Fix/file picker org assets root site

### LINKED ISSUE #1784: FilePicker: unable to browse through organization assets library when located on root site
#### Category
[ ] Enhancement

[X] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [3.17.0]

I have an **organization assets library** on the **root site** of my tenant.
I use the **FilePicker** component to browse inside this specific library via the relevant option.

#### Observed Behavior
When I select my organization assets library on root site, I have an infinite loading:
image
(sorry for the french texts).

The error is:
image

I think this is because of this following request to get the items of my library:
https://\*\*\*.sharepoint.com/sites/my-site/_api/SP.List.GetListDataAsStream?listFullUrl=%27https://\*\*\*.sharepoint.com//%27&RootFolder=/Org%20Assets

We can see a double slash after **.sharepoint.com**: //

#### Steps to Reproduce
1. In PowerShell, specify an organization assets library located on the root site of the tenant
2. Add a FilePicker component to your webpart
3. Browse inside the option "Your organization" (as I am french, I don't know the correct translated text for the option).

albegut:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #1784 

Fixed bug 1784. It is now possible to use the file picker independently of the location of the libraries.

### DIFF
--- a/src/services/OrgAssetsService.ts
+++ b/src/services/OrgAssetsService.ts
@@ -10,6 +10,17 @@ export class OrgAssetsService extends FileBrowserService {
   // Site organization assets library server relative URL
   private _orgAssetsLibraryServerRelativeSiteUrl: string = null;
 
+  private get orgAssetsLibraryServerRelativeSiteUrl(): string {
+    if (this._orgAssetsLibraryServerRelativeSiteUrl === "/") {
+      return "";
+    }
+    return this._orgAssetsLibraryServerRelativeSiteUrl?.replace(/\/$/, "") ?? null;
+  }
+
+  private set orgAssetsLibraryServerRelativeSiteUrl(value: string) {
+    this._orgAssetsLibraryServerRelativeSiteUrl = value;
+  }
+
   /**
    * Constructor
    * @param context Component context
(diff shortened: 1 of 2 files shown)


src/services/OrgAssetsService.ts
```diff
@@ -10,6 +10,17 @@ export class OrgAssetsService extends FileBrowserService {
   // Site organization assets library server relative URL
   private _orgAssetsLibraryServerRelativeSiteUrl: string = null;
 
+  private get orgAssetsLibraryServerRelativeSiteUrl(): string {
+    if (this._orgAssetsLibraryServerRelativeSiteUrl === "/") {
```
AJIXuMuK, line 14: Would it make more sense to have this logic in the setter instead of getter?
In that case it will run less times.

AJIXuMuK (changes requested):
Thank you @albegut for the changes?
I have one small comment - would be nice if you could address it!

src/services/OrgAssetsService.ts
albegut: Sure, I'll change it as soon as I can

albegut: Updated!

Issue #1275, opened 2022-07-28
add support for modernTaxonomyPicker in dynamic form


ReuvenDan:
#### Category
[ ] Enhancement

[ X ] Bug

[ ] Question

#### Version
  3.6.0  but also in 3.9.0(current latest)
_If you are not using the latest release, please update and see if the issue is resolved before submitting an issue._

#### Expected / Desired Behavior / Question
In the  dynamic form conponent
taxonomy field will show as "modern taxonomy picker" and support RTL terms.
field will support multi level terms trees

#### Observed Behavior
taxonomy field uses "TaxonomyPicker" component that does not support RTL.
field does not support multi level terms trees

#### Steps to Reproduce
create new termSet with RTL language terms.
Add a taxonomy field to the list and connect to the term set created above.
open the dynamic form.

Thanks!
![image](https://user-images.githubusercontent.com/34939784/181505040-b9391e82-d9f4-44ba-8199-cf17098c913a.png)

NishkalankBezawada:
Hello @michaelmaillot I can work on this once #1688 issue is merged.  

//Nishkalank

joelfmrodrigues:
@NishkalankBezawada the PR for #1688 is merged and since you volunteered, I have assigned this item to you 😊
Many thanks!

michaelmaillot:
Hello @ReuvenDan,

FYI there's an enhancement provided by @NishkalankBezawada which add support for `ModernTaxonomyPicker` in the `DynamicForm` Control, available in the next beta release and planned for version 3.21.0.

PR #1962, opened 2025-03-04
💡Added support to ModernTaxonomyPicker in DynamicForm - Fix 1275 [Repo Rescuer Challenge]

NishkalankBezawada:
| Q               | A
| --------------- | ---
| Bug fix?        | [ ]
| New feature?    | [x]
| New sample?      | [ ]
| Related issues?  | enhances #1275 

Added support to `ModernTaxonomyPicker` control in Dynamic Form. This will be handled by a property `useModernTaxonomyPicker`

| Property | Type | Required | Description |
| ---- | ---- | ---- | ---- |
| useModernTaxonomyPicker | boolean | no | Specifies if the form should render ModernTaxonomyPicker control for Managed metadata fields. If set to true, Dynamic form will render ModernTaxonomyPicker control. If set to false, Dynamic form will render TaxonomyPicker control. Default is false |

By using this property, we ensure that both TaxonomyPicker control and ModernTaxonomyPicker control can be used. This will be automatically defaulted to use `TaxonomyPicker ` control.

Fix1275

#### Guidance @joelfmrodrigues 

Thanks,
Nish


michaelmaillot:
Hello Nish, thanks for providing this PR.

Here's my two cents about this one, to go further this update. At some point, I was thinking about definitely replacing `TaxonomyPicker` by the `ModernTaxonomyPicker` in this control. But for the moment, the `ModernTaxonomyPicker` relies on the PnPjs v2 @sp/taxonomy package which uses v2.1 endpoint under the hood. But we'll problably update to PnPjs v4 at some point and this will involve to migrate from SP REST call to Graph call (or keep v2.1 REST call but having to overwrite existing calls to make explicit REST API calls instead of using PnPjs taxonomy calls).

From my perspective, with this PR we can let users choose to use the picker they want. But once we plan to update PnPjs version, we'll have to keep in mind that this will involve code base / documentation update regarding the `ModernTaxonomyPicker`.

src/controls/dynamicForm/dynamicField/DynamicField.tsx
```diff
@@ -531,20 +551,39 @@ export class DynamicField extends React.Component<IDynamicFieldProps, IDynamicFi
...
+          {useModernTaxonomyPickerControl ?
+            <div className={styles.pickersContainer}>
+              <ModernTaxonomyPicker
+                label=""
+                disabled={disabled}
+                initialValues={valueToDisplay !== undefined ? [valueToDisplay] : defaultValue}
+                placeHolder={placeholder}
+                allowMultipleSelections={false}
+                termSetId={fieldTermSetId}
+                anchorTermId={fieldAnchorId}
```
michaelmaillot, line 563: Using this value on a list form example, I have the following issue when I search for a term in the appropriate field:

image

Result displayed:

image


This is due to `fieldAnchorId` which is set to "00000000-0000-0000-0000-000000000000". I'm wondering if this value has to be used in this case (but to be honest, I'm not using this control a lot).

Do you see a case where this could be used properly in a metadata field?

docs/documentation/docs/controls/DynamicForm.md
```diff
@@ -66,9 +66,10 @@ The `DynamicForm` can be configured with the following properties:
...
+| customIcons | { [ columnInternalName: string ]: string } | no | Specifies custom icons for the form. The key of this dictionary is the column internal name, the value is the Fluent UI icon name. |
 | storeLastActiveTab | boolean | no |  When uploading files: Specifies if last active tab will be stored after the Upload panel has been closed. Note: the value of selected tab is stored in the queryString hash. Default - `true` |
 | folderPath | string | no | Server relative or library relative folder to create the item in. This option is only available for document libraries and works only when the contentTypeId is specified and has a base type of type Document or Folder. Defaults to the root folder of the library. |
+| useModernTaxonomyPicker | boolean | no | Specifies if the form should render ModernTaxonomyPicker control for Managed metadata fields. If set to true, Dynamic form will render ModernTaxonomyPicker control. If set to false, Dynamic form will render TaxonomyPicker control. Default is false |
```
michaelmaillot, line 72:
```suggestion
| useModernTaxonomyPicker | boolean | no | Specifies if the form should render [ModernTaxonomyPicker](./ModernTaxonomyPicker.md) control for Managed metadata fields. If set to `true`, Dynamic form will render ModernTaxonomyPicker control. If set to `false`, Dynamic form will render TaxonomyPicker control. Default is `false` |
```

michaelmaillot:
Hey @NishkalankBezawada,

Nice improvment! I've added some feedback, please let me know if you need help.

Would you mind resolving conflicts also?

michaelmaillot (changes requested):
Hey @NishkalankBezawada,

Nice improvment! I've added some feedback, please let me know if you need help.

Would you mind resolving conflicts also?

src/controls/dynamicForm/dynamicField/DynamicField.tsx
NishkalankBezawada: Guess we need to check on ModernTaxonomy picker control, I have checked your scenario, and using TaxonomyPicker, we can search the term, but not with ModernTaxonomy picker.

NishkalankBezawada:
Hey @michaelmaillot 

All set for your review.

Thanks,
Nish

src/controls/dynamicForm/dynamicField/DynamicField.tsx
michaelmaillot: ModernTaxonomy control is working well. I tried it through the `ControlsTest` web part and no issues from the moment where I only set the `termSetId` value. If I set the `anchorTermId` to a term which has children, it also works.

In the DynamicForm case, I think there should be an additional step that checks if an anchor term is different from the "empty Guid" value and if so, has to be added to the component, otherwise set the anchor term prop to null.

.eslintrc.js
```diff
@@ -34,7 +34,7 @@ module.exports = {
         //
         // STANDARDIZED BY:   @typescript-eslint\eslint-plugin\dist\configs\recommended.json
         '@typescript-eslint/explicit-function-return-type': [
-          1,
+          0,
```
michaelmaillot, line 37: Please don't touch the linter rules. This should be addressed separately.

package.json
```diff
@@ -127,7 +127,7 @@
     "react-test-renderer": "17.0.1",
     "request-promise": "4.2.5",
     "sonarqube-scanner": "2.8.2",
-    "spfx-fast-serve-helpers": "1.20.2",
+    "spfx-fast-serve-helpers": "~1.20.0",
```
michaelmaillot, line 130: `spfx-fast-serve-helpers` version should be explicit.

src/controls/dynamicForm/dynamicField/DynamicField.tsx
NishkalankBezawada: Thanks @michaelmaillot. I haven't had time to check on this, was busy in SharePoint Hackathon. I'll check and do the changes tomorrow.!🫡🤘

michaelmaillot (changes requested):
Hey @NishkalankBezawada,

The error met when typing in the taxonomy field when using `ModernTaxonomyPicker` should be addressed. I've proposed an update as a suggestion to fix this.

Let me know if you need any assistance.

src/controls/dynamicForm/dynamicField/DynamicField.tsx
NishkalankBezawada: Added additional step to check if anchor term is different from the "emptyGuid", as suggested

.eslintrc.js
NishkalankBezawada: Reverted back

docs/documentation/docs/controls/DynamicForm.md
NishkalankBezawada: Changed as suggested

NishkalankBezawada:
> Hey @NishkalankBezawada,
> 
> The error met when typing in the taxonomy field when using `ModernTaxonomyPicker` should be addressed. I've proposed an update as a suggestion to fix this.
> 
> Let me know if you need any assistance.

Hello @michaelmaillot 

This is now been addressed. All set for your review.

Thanks,
Nish

package.json
michaelmaillot: Please rollback on this one.

src/controls/dynamicForm/dynamicField/DynamicField.tsx
```diff
@@ -560,7 +560,7 @@ export class DynamicField extends React.Component<IDynamicFieldProps, IDynamicFi
                 placeHolder={placeholder}
                 allowMultipleSelections={false}
                 termSetId={fieldTermSetId}
-                anchorTermId={fieldAnchorId}
+                anchorTermId={fieldAnchorId !== null ? fieldAnchorId : undefined}
```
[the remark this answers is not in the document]
NishkalankBezawada: In `buildFieldCollection` if there is no anchorterm set, then it is by default assigning empty guid string as the term value. So had to check that scenario and passed null if it is empty guid. And Yes, I have missed this in the DynamicField to default it to `fieldAnchorId`. As my check is making it redundant. I have removed it now.

NishkalankBezawada:
> `package.json` & `package-lock.json` are still in the PR 🙃
> 
> Could you please remove them?

Yepp, Its done now. 

//Nish

michaelmaillot:
Merged manually, thanks again for your contribution!

I've updated the PR to state it as an enhancement, not a bug fix according to me.

PR #1963, opened 2025-03-06
🐛 - Fix 1655- Dynamic form should not display any validation errors until the user clicks Save/Submit, or leaves the focus of a required control - (Repo Rescuer Challenge)

### LINKED ISSUE #1655: Dynamic Form Loads with Validation Errors if no List Item Id is set
[ ] Enhancement

[X ] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [3.15.0]

When loading a Dynamic Form with no list item id (new item), the form should not display any validation errors until the user clicks Save/Submit, or leaves the focus of a required control.

#### Observed Behavior
When loading a Dynamic Form with no list item id (new item) all of the required fields display validation errors by default.

#### Steps to Reproduce
Create A Dynamic Form with no list item id set, and link it to a list that has a required field.  You will see the errors load by default.

You can see this error reported by @sallaadithya in a sub comment for an existing bug here: https://github.com/pnp/sp-dev-fx-controls-react/issues/1586#issuecomment-1637459697

NishkalankBezawada:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #1655 

Fixing issue #1655 

#### Observation
When loading a Dynamic Form with no list item id (new item), the form should not display any validation errors until the user clicks Save/Submit, or leaves the focus of a required control.

#### Actual behavior
When loading a Dynamic Form with no list item id (new item) all of the required fields display validation errors by default.

#### Solution
In ```DynamicField.tsx``` component, individual field is set and the validation is done through the methods ```getRequiredErrorText()``` and ```getNumberErrorText()```

These methods currently are only checking with ```changedValue``` property, which will be always empty for a new item. An additional check is done on basis of ```listItemId``` assuming that the fact that for new item, this will be empty, we can achieve to not to show the validation errors on the form. 

#### Observed 

Issue1655

#### After change

Issue1615

#### Edit scenario

Issue1655-Edit

Thanks,
Nish

### DIFF
--- a/src/controls/dynamicForm/dynamicField/IDynamicFieldState.ts
+++ b/src/controls/dynamicForm/dynamicField/IDynamicFieldState.ts
@@ -3,4 +3,5 @@ export interface IDynamicFieldState {
   * The options available to the listPicker
   */
   changedValue: any; // eslint-disable-line @typescript-eslint/no-explicit-any
+  listItemId:any; // eslint-disable-line @typescript-eslint/no-explicit-any
 }
--- a/src/controls/dynamicForm/dynamicField/IDynamicFieldState.ts
+++ b/src/controls/dynamicForm/dynamicField/IDynamicFieldState.ts
@@ -3,5 +3,4 @@ export interface IDynamicFieldState {
   * The options available to the listPicker
   */
   changedValue: any; // eslint-disable-line @typescript-eslint/no-explicit-any
-  listItemId:any; // eslint-disable-line @typescript-eslint/no-explicit-any
 }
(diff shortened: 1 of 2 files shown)


src/controls/dynamicForm/dynamicField/IDynamicFieldState.ts
```diff
@@ -3,4 +3,5 @@ export interface IDynamicFieldState {
   * The options available to the listPicker
   */
   changedValue: any; // eslint-disable-line @typescript-eslint/no-explicit-any
+  listItemId:any; // eslint-disable-line @typescript-eslint/no-explicit-any
```
AJIXuMuK, line 6: Why do you need `listItemId` in `state`? Couldn't you just use it from `props`?

AJIXuMuK (changes requested):

src/controls/dynamicForm/dynamicField/IDynamicFieldState.ts
NishkalankBezawada: My bad, missed that. Changed as suggested :). 

//Nish

AJIXuMuK (approved):

AJIXuMuK:
Thank you @NishkalankBezawada for the change!

Issue #1961, opened 2025-03-04
[Dynamic Form] It will throw "a 'PrimitiveValue' or 'StartObject' node was expected" error when clean up a multiple lookup field


wuxiaojun514:
#### Category
[ ] Enhancement

[x] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [   3.20      ]

#### Observed Behavior
It will throw error when you clean up a LookupMulti field and save it in Dynamic Form. 

> Error making HttpClient request in queryable [400]  ::> {"odata.error":{"code":"-1, Microsoft.SharePoint.Client.InvalidClientQueryException","message":{"lang":"en-US","value":"An unexpected 'PrimitiveValue' node was found when reading from the JSON reader. A 'StartArray' node was expected."}}}

![Image](https://github.com/user-attachments/assets/dc4e1c79-43da-45ea-9f38-5685c786a724)

I think we should use `[]` instead of `null ` when save it in rest api
![Image](https://github.com/user-attachments/assets/4bed7d9f-26c6-4873-9be4-3f78a52088f6)

#### Steps to Reproduce
1.  Make sure your list has a multiple lookup field and create an item which has value on this multiple lookup field.
2. Create a form customization spfx solution with Dynamic Form control
3.  Edit the item with Dynamic Form and delete the value on multiple lookup field
4.  Click save button in the form, then you will get the exception.

This problem is similar as #1578 .
It should work if we change the `null `to `[]` . I can find sometime to fix it.

![Image](https://github.com/user-attachments/assets/57e83e09-33da-43ba-9f7b-a67f2a4cb35d)

wuxiaojun514:
Got it, I will create a pull request tomorrow

wuxiaojun514:
Hi @michaelmaillot ,

I have raised the PR to fix it, please take a look.

michaelmaillot:
Linked PR has been merged and should be available in next beta release.

Thanks for your contribution!

PR #1967, opened 2025-03-19
Hotfix for 1937 and 1961

### LINKED ISSUE #1937: [Dynamic Form] Cannot Clean Up Date Field and Save it
### Category
[ ] Enhancement
[X] Bug
[ ] Question

### Version
3.20.0

### Problem Detail
I want to clean up the date field in **Dynamic Form** to set value to null.
By default, Dynamic Form doesn't support text input on date field, you need overwrite this field and enable **_allowTextInput_** property on `DatePicker` control so you can type/clean the date on it

Image

It worked at least one year ago.

Recently my team found that they cannot clean up date and it will throw error like` Cannot read properties of null (reading 'toString')`

Image

The root casue is on `OnChange` method in [DynamicForm,tsx](https://github.com/pnp/sp-dev-fx-controls-react/blob/master/src/controls/dynamicForm/DynamicForm.tsx)

Image

However, if I modify the code from

` field.stringValue = newValue.toString();`

to
` field.stringValue = newValue?.toString();`

I still cannot save my cleanup, because its **newValue** is **null** and **Dynamic Form** will only add fields whose newValue is not null/undefined into changes.  So my cleanup on date fields will never be submitted.

Image

I will debug the Dynamic Form control deeply to understand the logic of this code.

### Steps to Reproduce

1. Make sure your list contains date field
2. Override this date field in Dynamic Form to make sure the **DateTimePicker** enabled `allowTextInput`  property
3. Test your form customizer with an existing list item and try to clean up this date field
4. You should get the same error as I got.

wuxiaojun514:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #1937 and #1961

This PR will fix the following issues:
1. #1937 Cannot clean up date field issue, and will enable `allowTextInput` property in date field control

2. #1961 Get exception when clean up a multiple lookup field


src/controls/dynamicForm/DynamicForm.tsx
```diff
@@ -459,7 +459,7 @@ export class DynamicFormBase extends React.Component<
         if (fieldcolumnInternalName.startsWith('_x') || fieldcolumnInternalName.startsWith('_')) {
           fieldcolumnInternalName = `OData_${fieldcolumnInternalName}`;
         }
-        if (field.newValue !== null && field.newValue !== undefined) {
+        if (field.newValue !== undefined) {
```
AJIXuMuK, line 462: hey @wuxiaojun514 - why are you removing the `null` check? what is the problem it causes?

wuxiaojun514: Hi @AJIXuMuK ,
Thanks for the review.
This is due to this check will cause #1937 . Since the value of cleaned date field is null, this check will stop adding cleaned date field into changes. Then the clean up of date will not be saved.
You can see my previous comment at https://github.com/pnp/sp-dev-fx-controls-react/issues/1937#issuecomment-2608720195 for more information

michaelmaillot:
Merged manually, thank you!

Issue #2061, opened 2025-10-19
Tab key in DynamicForm deletes field values


CodingSinceThe80s:
#### Category
[ ] Enhancement

[x ] Bug

[ ] Question

#### Version

3.22 (latest at time of this posting)

#### Expected / Desired Behavior 

After saving a form, I expect to be have the values persisted when editing the form afterwards. I want to switch through the fields with the Tab key and NOT lose the value of those fields when pressing the Tab key.

#### Observed Behavior
When using the Edit form (PageType=6) and Tab-ing through the fields, the field values are being cleared (I do not press any other key or move the mouse). Using the Tab key to navigate through a form is a common task and should not lead to deleting values from the form.

#### Steps to Reproduce

1. Create a custom list ( I used the fields Title, abcTotalAmount as number with 0 trailing digits and mandatory field, abcTaxAmount as number with 0 trailing digits and abcComments as multiple lines of text)
2. Create a spfx project (I used version 1.21) - form application extension. 
3. In the component use the following code:
```
import { DynamicForm } from "@pnp/spfx-controls-react/lib/DynamicForm";

export interface IMyFormsProps {
  context: any;
  displayMode: FormDisplayMode;
  targetListId: string;
  targetItemId: number;
  onSave: () => void;
  onClose: () => void;
}

//...component boilerplate omitted for brevity

public render(): React.ReactElement<{}> {

    return( <DynamicForm 
          context={this.props.context} 
          listId={this.props.targetListId}
          listItemId={ this.props.targetItemId != null ? this.props.targetItemId : 0 }
          contentTypeId={"0x01006CAD569..."}  //this is the list item content type for the custom list above 
          onCancelled={() => { console.log('Cancelled') }}
          onBeforeSubmit={async (listItem) => { return false; }}
          onSubmitError={(listItem, error) => { alert(error.message); }}
          onSubmitted={async (listItemData) => { console.log(listItemData); this.props.onSave(); }}
          >
        </DynamicForm>);
```
Also in MyFormCustomizer.ts:
```
React.createElement(MyForms, {
        context: this.context,
        displayMode: this.displayMode,
        targetItemId: this.context.itemId,
        targetListId: this.context.list.guid.toString(),
        targetContentTypeId: targetContentTypeId,
        onSave: this._onSave,
        onClose: this._onClose
       } as MyFormsProps)
```
4. Activate the extension for the custom list
5. Create a new item using the New Form
6. Edit the newly created item
[...]

PR #2070, opened 2025-11-19
Removed obsolete componentDidUpdate in DynamicField and changed onBlur

### LINKED ISSUE #1923: Dynamic form text, textarea and number fields clear values on blur after another field has been edited.
#### Category

[A dynamic form for a list with text, textarea, and number field inputs. loses values in those fields if you click or tab into the field and then click or tab out. This happens after you have edited another field. Or sometimes w/o even editing another field. This happens with no field overrides or other changes from the OOTB dynamic form. ] Bug

#### Version
I just updated to 3.20.0 and it's still happening

Please specify what version of the library you are using: [  3.20.0      ]

Expected behavior is for fields with values to retain those values on blur._

#### Observed Behavior
Observed behavior is text, textarea, and number fields losing value on blur.

#### Steps to Reproduce
Create a FormCustomizer and Dynamic form for an existing list with text, textarea, and number fields, and other fields as desired. 
gulp serve the edit form and click in and out of a text field. It may retain its value, however, next edit another field and then click or tab into any text, textarea, or number field and then click or tab out of the field. The behavior we're seeing is the field losing focus will then also lose its value. Saving the form will update the input field as empty in the list.

CodingSinceThe80s:
| Q               | A
| --------------- | ---
| Bug fix?        | [ x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #1923, mentioned in #2061 

Removed the componentDidUpdate method as suggested by t0mgerman in the issue discussion for 1923 and adjusted the onBlur method in DynamicField.tsx. Afterwards I tested this on a form with number, text, person and taxonomy fields - the issue seems to be correctly resolved, but please also test on your side to double check.

Also some formatting, but feel free to leave that out ;)


joelfmrodrigues:
@CodingSinceThe80s I reverted the gitignore change on the PR, but still left with a large volume of files and they are all from the `lib` folder who should not be included...
closing the PR as it should be easier and simpler to open a separate PR with the fix

PR #2099, opened 2026-03-07
#1332 - fixed DateTimePicker "showClearDate" button disabled state

### LINKED ISSUE #1332: DateTimePicker "showClearDate" button not disabled when component is set as disabled
#### Category
[ ] Enhancement

[X] Bug

[X] Question

#### Version

Please specify what version of the library you are using: [ 3.11.0 ]

Clear date button should be disabled too when the DateTimePicker is set as disabled.

image

I don't know for sure if this is the desired behavior since I can basically "Not show" the button when I want to disable the component. But in my opinion it makes more sense to make them both disabled with the same prop.

image

Ateina:
| Q               | A
| --------------- | ---
| Bug fix?        | [X ]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #1332

Clear date button is now hidden when the DateTimePicker is disabled, since clearing a date has no function in that state.

### DIFF
--- a/src/controls/dateTimePicker/DateTimePicker.tsx
+++ b/src/controls/dateTimePicker/DateTimePicker.tsx
@@ -368,7 +368,7 @@ export class DateTimePicker extends React.Component<IDateTimePickerProps, IDateT
                 }}
               />
             </div>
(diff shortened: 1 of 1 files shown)


AJIXuMuK:
Thank you @Ateina for the fix!

PR #2102, opened 2026-03-30
PeoplePicker: defaultSelectedUsers issue resolving groups with similar names. 

### LINKED ISSUE #2101: PeoplePicker: defaultSelectedUsers resolves wrong group when groups have similar names
#### Category
[ ] Enhancement

[ X] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [    3.25.0     ]

When people picker is used in a controlled way, for example in a new/edit form, defaultSelectedUsers should resolve the correct group.

#### Observed Behavior
When using defaultSelectedUsers and passing a group that has similar named group, the wrong group name is resolved. For example, I have 2 security groups, Communications and Communications Manager. If I set defaultSelectedUsers to the Communications group, when the component loads, Communication Manager is what is resolved.

Image

Image

Also, if I have defaultSelectedUsers pointed to a state variable, in an edit form expereince, for example, if I type Communications into the people picker and select it, defaultSelectedUsers is also called since there was a state update, and resolves it to the incorrect Communications Manager group.

I've traced this to the searchPersonByEmailOrLogin function in PeopleSearchService, which is hard-coded to only return 1 result, which works fine with email addresses or login names, but when dealing with security groups, we can only use the group name (as far as I can tell). 

Image

Using same exmaple, defaultSelectedUsers=["Communications"], this is what happens:

Communications is the query
Image

But the 1 result that is resolved is Communications Manager

Image

I am able to resolve this issue for both groups and users by increasing the max suggestions from 1 to a higher number. Currently, I am optionally using the maximumSuggestions prop and defaulting to 5 at a minimum. Open to suggestions on best way to configure that value, but doing this alleviates this issue:

Image

Same Communications value passed in

Image

Now I get both Communications and Communications Manager groups

Image

The userResult[0], in all of my groups testing contains the correct group.

Image

A better approach would be to filter this array based on the value that was passed in to ensure you get the proper result.

Image

I will submit a PR and can help make any proposed tweaks, but in my testing this approach has corrected the issue.

#### Steps to Reproduce
This can be tricky, the communications example above is present in my client's tenant, but if I create those two named groups in my tenant, it does not present. However, in my tenant, I have two groups, "English Test Group" and "English Test Group2" which repros the issue; when defaulting "English Test Group" the other group "English Test Group2" resolves when defaultSelectedUsers is called.

c-eiser13:
| Q               | A
| --------------- | ---
| Bug fix?        | [ X]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #2101 

This PR increases the results returned from the searchPersonByEmailOrLogin function that is called by the defaultSelectedUsers prop in the People Picker control. This function was previously hard-coded to return 1 result, which works well when using an email address or login to perform the search, but when a group is used and searched by name, there are instances where groups with a similar name are resolved incorrectly. Increasing the number of results, then filtering the array with the actual query value produces the correct resolved groups. See #2101  for more information. If any adjustments should be made, please let me know and I would be happy to do so!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -227,10 +227,10 @@
       "dev": true
     },
     "node_modules/@azure/core-http": {
-      "version": "3.0.4",
-      "resolved": "https://registry.npmjs.org/@azure/core-http/-/core-http-3.0.4.tgz",
-      "integrity": "sha512-Fok9VVhMdxAFOtqiiAtg74fL0UJkt0z3D+ouUUxcRLzZNBioPRAMJFVxiWoJljYpXsRi4GDQHzQHDc9AiYaIUQ==",
-      "deprecated": "This package is no longer supported. Please migrate to use @azure/core-rest-pipeline",
+      "version": "3.0.5",
+      "resolved": "https://registry.npmjs.org/@azure/core-http/-/core-http-3.0.5.tgz",
+      "integrity": "sha512-T8r2q/c3DxNu6mEJfPuJtptUVqwchxzjj32gKcnMi06rdiVONS9rar7kT9T2Am+XvER7uOzpsP79WsqNbdgdWg==",
+      "deprecated": "This package is no longer supported. Please refer to https://github.com/Azure/azure-sdk-for-js/blob/490ce4dfc5b98ba290dee3b33a6d0876c5f138e2/sdk/core/README.md",
       "dev": true,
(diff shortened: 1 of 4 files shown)


AJIXuMuK:
thank you @c-eiser13 for the fix!

PR #2126, opened 2026-07-27
Add aria-pressed to RichText text formatting buttons

### LINKED ISSUE #1945: Accessibility Issue in RichText: Screen reader doesn't announce state of text formatting controls whether its selected or not when accessing it through tab.
User Experience:
Screen reader dependent users will be impacted as the Selected state is not defined for Style buttons. Due to this, users will not be able to identify the state of the control. This will confuse them.

Actual Result:
Selected state is not defined for Style buttons. i.e. Bold, Italic, Underline.

Note:
Same issue repro if we unselect button, screen reader doesn't announce any information regarding it.

Expected Result:
Selected state should be defined for Style buttons.
Screen reader should announce "selected" information when a button is selected.

Please see attached screenshot where i have selected Underline button but it doesn't announce whether it has been selected or not and also as you can see bold was already selected earlier and when i navigate to it by pressing tab it doesn't announce the selected state of that button which will confuse screen reader users.

Image

jtlivio:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?     | [ ]
| Related issues? | fixes #1945

The Bold, Italic, Underline, and Link buttons pass `checked` to Fluent UI's
`IconButton` without `toggle`. `aria-pressed` is only rendered when `toggle`
is set, so screen readers never announce whether a format is active. Since
`checked` reflects the active formatting state at the current selection,
adding `toggle` ensures that `aria-pressed` is exposed correctly for all four
buttons.

Added regression coverage for Bold, Italic, Underline, and Link, verifying that
each button exposes `aria-pressed="false"` in its inactive state.

The new tests fail without the fix. `npm run build` passes with no new warnings.

### DIFF
--- a/src/controls/richText/RichText.test.tsx
+++ b/src/controls/richText/RichText.test.tsx
@@ -1,0 +1,60 @@
+import * as React from 'react';
+import { expect } from 'chai';
+import { mount, ReactWrapper } from 'enzyme';
+import { RichText } from './RichText';
+
+describe('<RichText />', () => {
+  let host: HTMLDivElement;
+  let richtext: ReactWrapper;
+
+  // The toolbar positions itself from its parent element on mount, so the
+  // component has to be attached to the document rather than mounted detached.
+  beforeEach(() => {
+    host = document.createElement('div');
+    document.body.appendChild(host);
+  });
+
+  afterEach(() => {
+    richtext.detach();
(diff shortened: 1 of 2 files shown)


joelfmrodrigues:
@jtlivio many thanks! Merged

PR #2157, opened 2026-09-21
hotfix for #2156 (DynamicForm hide/show formula doesn't work issue)

### LINKED ISSUE #2156: [DynamicForm] hide/show conditional formula doesn't work because it wrongly handles string false values
#### Category
[ ] Enhancement

[x] Bug

[ ] Question

#### Version

Please specify what version of the library you are using: [  3.25      ]

`DynamicForm` supports sharepoint form [conditional hide/show formulas](https://learn.microsoft.com/en-us/sharepoint/dev/declarative-customization/list-form-conditional-show-hide?source=recommendations) (You can configure a rule to hide field B when field A's value is in some condition)
But I found it doesn't work when I upgrade to 3.25

I remembered that it worked in 3.22 or 3.23 

Image

#### Observed Behavior

Right now `DynamicForm` won't hide the field even when it match the conditional formula rule

It seems that the root cause is in https://github.com/pnp/sp-dev-fx-controls-react/blob/dev/src/controls/dynamicForm/DynamicForm.tsx#L951
```ts
      if (Boolean(result) !== true) {
          results[fieldName] = message;
        }
```
The `Boolean(result)` always return true even the `result` is `'false'`

It works as expected if I change this line to
```ts
if (result !== true && result !== "true") {
          results[fieldName] = message;
        }
```

I can try to raise a PR to fix it

wuxiaojun514:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | [ ]
| New sample?      | [ ]
| Related issues?  | fixes #2156 

Currently the conditional hide/show formula doesn't work in `Dynamic Form` .
This is due to `if (Boolean(result) !== true) {` check doesn't work properly in  https://github.com/pnp/sp-dev-fx-controls-react/blob/dev/src/controls/dynamicForm/DynamicForm.tsx#L951 
If the `result` variable is `'false'`, `Boolean(result)` will be `true`. So it will never hide the related column.


joelfmrodrigues:
@wuxiaojun514 many thanks!

